Issue #10, opened 2015-07-28
Add support for python 2


relekang:

relekang:
Releated: #9

relekang:
Benefit: be to be able to use the setup.py hook in python 2 projects

relekang:
[Should avoid this](https://ci.frigg.io/relekang/django-nopassword/70/)

jezdez:
Yeah, I'd vote for Python 2 support. I'm happy to provide a patch if you'd like.

relekang:
Yeah, that would be great.

jezdez:
Cool! Will do this week :)

relekang:
Sweet, I think a27e282 removes all usage of python 3 only features from configparser :)

PR #14, opened 2015-08-03
Fixed name of assertion function.

jezdez:
Seems like this snuck in and tested for the wrong thing. The assertion function
is called assert_called_once_with(), only running assert_called_once() will
call the mock instead.

Refs #10.


jezdez:
FTR this actually fails now for me :(

jezdez:
Aha, that fixed it.

relekang:
Nice catch, thanks :)

PR #52, opened 2016-01-26
Add "feature" as an alternative to "feat" in the parser.

skorokithakis:
I always forget and write `feature` rather than `feat`, so I figured I'd add it as an alias in the angular parser.

### DIFF
--- a/semantic_release/history/parser_angular.py
+++ b/semantic_release/history/parser_angular.py
@@ -4,7 +4,7 @@ from ..errors import UnknownCommitMessageStyleError
 from .parser_helpers import parse_text_block
 
 re_parser = re.compile(
-    r'(?P<type>feat|fix|docs|style|refactor|test|chore)'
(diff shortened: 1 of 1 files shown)


skorokithakis:
Can this be merged?

relekang:
Sorry for the late answer. I think this should be opt-in somehow, either by being a separate parser that extends the angular parser or by being an option.

skorokithakis:
Hmm, why do you say that? Is there a case where someone would start a commit with "feature:" but didn't want it to count as a feature?

relekang:
As it is called the angular parser it should fit with the specification in the angular commit guidelines. This is important to make it possible to work with other tools. I am not against extending it as long as it follows the specifications as default. Furthermore, consistency is good so I rather see it enforce one label for each type as default.

relekang:
>  I'm not sure how to make this an option, though.
I think the easies way would be to add another parser.

relekang:
I'm closing this, if you want this feature to please open a pr with a new parser.

PR #54, opened 2016-03-13
add option to use tag instead of creating new commit when versioning

KenMercusLai:
I changed the behaviours of finding current version and versioning

1. original way was using configed file to track version, I think using tag would be more straight forward. it returns 0.0.0 as default when it fails to find any version tag
2. instead of changing file content, I simply use tag to tract commits. I'm not sure this is a good way of versioning since I'm not a professional programmer.

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -62,18 +62,13 @@ def version(**kwargs):
         if not check_build_status(owner, name, get_current_head_hash()):
             click.echo(click.style('The build has failed', 'red'))
             return False
-        click.echo(
-            click.style('The build was a success, continuing the release', 'green'))
+        click.echo(click.style('The build was a success, continuing the release', 'green'))
 
-    # the default behavour for versioning is commit
-    # using tag when it is explicitly configed
-    if not (config.has_option('semantic_release', 'versioning_by_tag') and
-            config.getboolean('semantic_release', 'versioning_by_tag')):
+    if config.get('semantic_release', 'version_source') == 'commit':
         set_new_version(new_version)
         commit_new_version(new_version)
     tag_new_version(new_version)
-    click.echo(
-        'Bumping with a {0} version to {1}.'.format(level_bump, new_version))
+    click.echo('Bumping with a {0} version to {1}.'.format(level_bump, new_version))
     return True
 
 
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -64,8 +64,8 @@ def version(**kwargs):
             return False
         click.echo(click.style('The build was a success, continuing the release', 'green'))
 
-    set_new_version(new_version)
-    commit_new_version(new_version)
+    # set_new_version(new_version)
+    # commit_new_version(new_version)
     tag_new_version(new_version)
     click.echo('Bumping with a {0} version to {1}.'.format(level_bump, new_version))
     return True
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -64,8 +64,12 @@ def version(**kwargs):
             return False
         click.echo(click.style('The build was a success, continuing the release', 'green'))
 
-    # set_new_version(new_version)
-    # commit_new_version(new_version)
+    # the default behavour for versioning is commit
+    # using tag when it is explicitly configed
+    if not (config.has_option('semantic_release', 'versioning_by_tag')
+        and config.getboolean('semantic_release', 'versioning_by_tag')):
+        set_new_version(new_version)
+        commit_new_version(new_version)
     tag_new_version(new_version)
     click.echo('Bumping with a {0} version to {1}.'.format(level_bump, new_version))
     return True
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -62,16 +62,18 @@ def version(**kwargs):
         if not check_build_status(owner, name, get_current_head_hash()):
             click.echo(click.style('The build has failed', 'red'))
             return False
-        click.echo(click.style('The build was a success, continuing the release', 'green'))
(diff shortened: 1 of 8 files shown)


relekang:
Thanks you for submitting this. Why do you think this is a better approach?

KenMercusLai:
Simply because it won’t touch any existed file & create a new commit for indicating a new version.

A tag for me is an additional marker to a commit, so I think this could keep git history only focuses on changes.

relekang:
Makes sense. How would this work with a package, e.g. a tarball, that does not contain the git history? If you unpack and run `python setup.py <command>` will it then default to `'0.0.0'`?

I guess it would be nice to have the choice between the two approaches, can you change your PR to add this as an option? With the old as default, I would like to test this a bit before considering changing the default behaviour.

haeric:
Seems like a great option to have to work with tags only, as long as it's an option and not mandatory :)

KenMercusLai:
if a dir is not a git repo, this will simply fail because of init part of the code

sure, I'll start working on it as an option ASAP

KenMercusLai:
added a option for choosing versioning by commit or tag with the key: versioning_by_tag in the config file

it will version by using tag unless it is explicitly set True with versioning_by_tag

KenMercusLai:
@relekang I'm having some problem with the tests. like the repo owner, I have to change it to me to pass the local test, but this will cause the ci test fail.

How should I solve this?

relekang:
If the test is not related what you are working on leave it be :)

I think a few tests will be fixed if you rebase on master, could you please do that?

KenMercusLai:
seems everything works fine after rebasing.

for the question about package you mentioned, I think it won't contain any version info in the package source code since it will not be automatically added/updated.

but version info could be contained in the file specified for it. I think this won't cause any issue. But I'm not a professional programmer, I'm not sure about it.

KenMercusLai:
BTW, I think the doc is not very easy to understand for people like me. I tried days to setup travisci and other tool to work properly.

This tool is very useful if dev can setup a CI or workable environment.

semantic_release/cli.py
```diff
@@ -62,12 +62,18 @@ def version(**kwargs):
         if not check_build_status(owner, name, get_current_head_hash()):
             click.echo(click.style('The build has failed', 'red'))
             return False
-        click.echo(click.style('The build was a success, continuing the release', 'green'))
-
-    set_new_version(new_version)
-    commit_new_version(new_version)
+        click.echo(
```
relekang: Why did you add a newline here?

semantic_release/history/__init__.py
```diff
@@ -25,6 +35,12 @@ def get_current_version():
             re.MULTILINE
         ).group(1)
 
+if (config.has_option('semantic_release', 'versioning_by_tag') and
```
relekang: If you add the option to defaults.cfg it is not necessary to check if it exists. Can you please add it with false as default value and remove all `has_option` calls?

semantic_release/history/logs.py
```diff
@@ -32,17 +32,31 @@ def evaluate_version_bump(current_version, force=None):
...
-            changes.append(message[0])
-        except UnknownCommitMessageStyleError:
-            pass
-
-        commit_count += 1
+    if (config.has_option('semantic_release', 'versioning_by_tag') and
+            config.getboolean('semantic_release', 'versioning_by_tag')):
+        # we have to first find our the version hash tagged correspondingly
+        version_hash = get_version_from_tag('v{0}'.format(current_version))
+        for commit_message in get_commit_log(version_hash):
```
relekang: This is the same as `for commit_message in get_commit_log('v{0}'.format(current_version)):`, thus I think it is unnecessary to get the hash first. If you change this back, the only difference between tag and regular here is that the version will not end up in the commit message so the break won't be called for tag, but that is equal to the other case here. I think this can be changed back to the way it was and still work with the tag version.

relekang:
What about renaming the option to `version_source` with `variable` and `tag` as options and add `version_source=variable` to defaults.cfg?

relekang:
> seems everything works fine after rebasing.

:+1: 


> BTW, I think the doc is not very easy to understand for people like me. I tried days to setup travisci and other tool to work properly.

Are you talking about this http://python-semantic-release.readthedocs.org/en/latest/automatic-releases/travis.html ?

KenMercusLai:
@relekang I've made changes. But I'm facing a problem in writing a test. 
Since using version_source as a string, I need to patch the configparser.get to return specific value some version_source while leave other untouched.

tests/test_cli.py
```diff
@@ -41,12 +41,11 @@ def test_version_by_commit_should_call_correct_functions(mocker, runner):
...
-            return True
-        return False
-    mocker.patch('semantic_release.cli.config.getboolean', config_file_path)
-    mocker.patch('semantic_release.cli.config.has_option', lambda *x: True)
+    # def config_file_path(*args):
+    #     if args[0] == 'semantic_release' and args[1] == 'version_source':
+    #         return True
+    #     return False
+    # mocker.patch('semantic_release.cli.config.getboolean', config_file_path)
     mock_tag_new_version = mocker.patch('semantic_release.cli.tag_new_version')
```
KenMercusLai: how can i patch here to have confi.get return 'tag' when the args are 'semantic_release' and 'version_source' while leave return of other parameters alone

KenMercusLai:
@relekang After a long absence, I finally have time to finish these tests. Can you review it?
😸

relekang:
I hope to have time in the next few days. Thanks for your patience ✌️

relekang:
LGTM, thanks 👍

PR #62, opened 2016-04-17
fix: Add search_parent_directories option to gitpython

relekang:


relekang:
@salimfadhley I think this might solve the problem mentioned in the email.

salimfadhley:
Excellent, I will try it out when I'm at work tomorrow.

PR #69, opened 2017-06-30
Update Twine

honzajavorek:
The publishing API is under development and older versions of Twine have problems to deal with newer versions of the API. Namely the logic of register/upload has changed (it was simplified).

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,6 +2,6 @@ click==6.2
 gitpython==1.0.1
 invoke==0.11.1
 semver==2.2.1
-twine==1.5.0
+twine==1.9.1
 requests==2.9.1
 wheel


relekang:
Thanks, sorry for missing this for so long 😅

relekang:
Oh, the CI used in this repo is shutdown. I will setup circle later so we can get it released.

argos83:
@honzajavorek @relekang  with this version of twine (1.9.1, the latest at the moment) I get this error: 

```
semantic_release/pypi.pyc in upload_to_pypi(dists, username, password)
     18         password=password,
     19         comment=None,
---> 20         sign_with='gpg'
     21     )
     22     run('rm -rf build dist')

TypeError: upload() takes exactly 13 arguments (8 given)
```
With python 2.7 and semantic-release 3.10.1

Does it work for you guys? I guess there was a reason why twine was pinned to an old version (which no longer works since it uses a no longer supported pypi API)

relekang:
>That was merged already, but not published. @relekang could you publish it?

That PR was not merged. 

Latest master is released so if the problem is in the release it is on current master. Would you creating a PR with the new arguments? I won't have time to do it this week.

argos83:
There's the PR I would really appreciate if you could merge it and release a new version any time soon. Thanks!

PR #70, opened 2017-07-22
fix: Fix broken uploading code

skorokithakis:
This fixes broken uploading due to the new PyPI API by upgrading Twine to the latest version.

It also reorders operations so that PyPI uploading is the last step, as it's the easiest to then do manually if it breaks again.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,6 +2,6 @@ click==6.2
 gitpython==1.0.1
 invoke==0.11.1
 semver==2.2.1
-twine==1.5.0
+twine==1.9.1
 requests==2.9.1
 wheel
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -125,12 +125,6 @@ def publish(**kwargs):
             name=name
         )
 
-        if config.getboolean('semantic_release', 'upload_to_pypi'):
-            upload_to_pypi(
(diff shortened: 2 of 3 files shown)


relekang:
Thanks :v:

I am a bit unsure about doing the changelog posting before it is uploaded. It will notify all watchers of a new release, thus it would not be good to post unless the publishing is successful? There is a command in the cli for posting the changelog of the last release so it should be ok to do manually.

relekang:
Just realised #69 existed so I merged that since it did not contain the reordering.

skorokithakis:
Oh changelog posts the changelog to GitHub? I thought it only generated it?

If the release doesn't succeed, I just do it manually, but if the tool works and the changelog command posts the changelog, that's enough for me, thank you.

skorokithakis:
Oh, have/can you make a new release?

relekang:
Yeah, just realised that this is still setup with frigg, which has shut down. Will setup circle and make sure it releases later tonight.

relekang:
>Oh changelog posts the changelog to GitHub? I thought it only generated it?

Yes, it posts to github if it has the environment variable GH_TOKEN or is on one of the supported CI's that uses another env variable

skorokithakis:
Fantastic, thanks!

PR #75, opened 2018-01-27
Fix: error when not in git repository

blackHatMonkey:
Fix an error when the program was run in a non-git repository. It would not allow the help options to be run.

@relekang I know you wanted to put this `repo = Repo('.git', search_parent_directories=True)` line in it's own function. However that just adds unnecessary overhead and furthermore it does not lead to cleaner code.


relekang:
Thanks 🙂

PR #78, opened 2018-04-10
Add --retry cli option

skorokithakis:
Many times, `semantic-release` will die with an error (because of missing tags, files existing on PyPI because I forgot to delete the `dist/` dir, etc etc), leaving me with a half-made commit that I basically have to resume by hand.

This PR makes `semantic-release` idempotent by adding a `--retry` option that will retry the same commit, and will not change the repository. It will also skip existing files on PyPI so the release will be much more likely to succeed.


relekang:
Master has been updated with CI integration, do you mind rebasing on master so CI will build?

skorokithakis:
Not at all, I'll do that in a few minutes, thank you.

relekang:
Also please add the --retry option to the documentation.

skorokithakis:
This should be done!

relekang:
Thanks :v:

relekang:
It seems python2 broke after one of the recent pull-requests I hope to have a look at it soon and then do a release.

relekang:
This have been released in v3.11.0

skorokithakis:
Fantastic, thank you!

Issue #80, opened 2018-05-13
GitPython locked to old version causes PackingType exception


kevinkjt2000:
My first issue here, so I wanted to say thanks for bringing the loveliness of semantic release to python!

Today I ran into an issue with GitPython that has already been fixed in a newer version https://github.com/gitpython-developers/GitPython/issues/687

It seems that the requirements.txt file here is locking GitPython to an old version.  Any chance that could be updated?

relekang:
Yeah, sure I might be able to do this in the end of the week. If you want it faster a PR is always welcome. :v:

relekang:
Fixed in v3.11.1 thanks to @blackHatMonkey

PR #82, opened 2018-06-06
fix: change Gitpython version number

blackHatMonkey:
Change the Gitpython version number to fix a bug described in #80.


relekang:
Thanks 🙌

PR #83, opened 2018-06-09
chore: remove python2 support

blackHatMonkey:
This patch removes the `python2` support for project. Going forward it will be only `python3` compatible which provides multiple advantages such as cleaner syntax, etc. Also this significantly improves the `tox` testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,40 +1,6 @@
 version: 2
 
 jobs:
-  flake8:
-    docker:
-      - image: circleci/python:3.6
-
-    steps:
-      - checkout
-      - run: sudo pip install tox
-      - restore_cache:
-         keys:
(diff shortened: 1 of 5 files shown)


relekang:
This removes a lot from the tox setup unrelated to python2 that where put there for a reason. That said, people put there time and effort into contributed the python 2 support so they could use this library. Furthermore, it is not in our way so we are not going to throw it without a good reason. Please open an issue explaining something that can't be done while keeping python 2 support if you still think this is necessary.

blackHatMonkey:
@relekang `python2` will only be supported for about [a year and a half](https://pythonclock.org/). By dropping `python2` support it will make dependencies management much easier. Even by dropping `python2` support users of `python2` will still be able to download the tool but it's not gonna be the latest version.

That being said if you still want to keep `python2` support there is the `future` library which allows for monkey patching the standard library and should make it easier to transition to `python3` when it is needed. Let me know what you think of this.

PR #87, opened 2018-09-03
fix: Never uploading existing to pypi

### LINKED ISSUE #86: Publish to PyPI attempts to upload old distributions
I'm getting an error where a new release is being generated properly and is uploaded to PyPI, but 
the CircleCI build step fails because python-semantic-release also attempts to upload older distributions to PyPI. 

The relevant step in the CircleCi workflow: 
```yaml
- run:
      name: upload to pypi
      command: |
          git config --global user.email "aumitleon1@gmail.com"
          git config --global user.name "aumitleon"
          . venv/bin/activate
          semantic-release publish
```

After successfully uploading to the newest version to PyPI (v0.2.5), it also attempts to upload the previous version (v0.2.4), which causes an error:
screen shot 2018-09-03 at 2 42 59 pm

I suspect the issue has to do with setting `skip_existing` to the `retry` boolean:
https://github.com/relekang/python-semantic-release/blob/762fbcfb72f3ed269eaa1bbb8b0de433166a0a47/semantic_release/cli.py#L156

Is there a reason why `skip_existing` is set to retry? 

From my understanding, the `retry` option just attempts to publish changes using the existing version, without bumping a new version. I don't think the `skip_existing` option for twine should be tied to this boolean because `skip_existing` is meant to avoid uploading any distributions already on PyPI (See: https://github.com/pypa/twine/issues/63), which I believe is the cause of this error. 

I've submitted a PR that I think should resolve this issue. Open to any comments or questions 😃

AumitLeon:
This aims to to resolve #86. Previously, the `retry` boolean flag was directly tied to the `skip-existing` flag within twine upload step for uploading to PyPI. I believe the previous rationale for this is captured by this comment: 
https://github.com/relekang/python-semantic-release/blob/762fbcfb72f3ed269eaa1bbb8b0de433166a0a47/semantic_release/cli.py#L155

Tying `retry` to `skip_existing` misses the case where a user attempts to publish to PyPI without specifying the `retry` flag.

With these new changes, when uploading to PyPI, we will always skip attempting to upload a distribution already hosted on PyPI (which throws an error). 

If a user wants to specify the `retry` flag when attempting to upload to PyPI, then `skip_existing` is already set to true by default, so we won't attempt to upload existing files. If a user _doesn't_ explicitly specify the `retry` flag, then this change ensures that we don't attempt to upload existing files to PyPI. 

Basically, when uploading to PyPI, I don't think we ever want to upload files that already exist on PyPI.

Also, from Twine's description of the `skip_existing` flag: 

> Specify whether twine should continue uploading files if one of them already exists. This primarily supports PyPI. Other package indexes may not be supported.

So it seems like it makes sense for `skip_existing` to default to `True` when uploading to PyPI.

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -151,9 +151,7 @@ def publish(**kwargs):
         if config.getboolean('semantic_release', 'upload_to_pypi'):
             upload_to_pypi(
                 username=os.environ.get('PYPI_USERNAME'),
-                password=os.environ.get('PYPI_PASSWORD'),
-                # We are retrying, so we don't want errors for files that are already on PyPI.
-                skip_existing=retry,
+                password=os.environ.get('PYPI_PASSWORD')
             )
 
         if check_token():
--- a/semantic_release/pypi.py
+++ b/semantic_release/pypi.py
@@ -2,7 +2,7 @@ from invoke import run
 from twine.commands import upload as twine_upload
 
 
-def upload_to_pypi(dists='sdist bdist_wheel', username=None, password=None, skip_existing=False):
+def upload_to_pypi(dists='sdist bdist_wheel', username=None, password=None):
     """
     Creates the wheel and uploads to pypi with twine.
 
@@ -19,7 +19,7 @@ def upload_to_pypi(dists='sdist bdist_wheel', username=None, password=None, skip
         comment=None,
         sign_with='gpg',
         config_file='~/.pypirc',
-        skip_existing=skip_existing,
+        skip_existing=True,
         cert=None,
         client_cert=None,
         repository_url=None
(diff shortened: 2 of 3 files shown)


AumitLeon:
Looks like the `test_upload_without_arguments` is failing. I think this is because it's asserting that skip-existing is false:
https://github.com/relekang/python-semantic-release/blob/762fbcfb72f3ed269eaa1bbb8b0de433166a0a47/tests/test_pypi.py#L17-L31

The changes in this PR would like to set `skip-existing` to `True` when uploading to PyPI.

AumitLeon:
Is there anything I should change about this PR? Open to any thoughts! Thanks for the opportunity to contribute! 😃

relekang:
The problem should be fixed now since the dist folder is cleaned before building for upload. Then it will still throw if we the recent build is an existing version. Unless retry is passed this is a situation that should not fail silently.

PR #89, opened 2018-09-29
fix: initialize git Repo from current folder

splendido:
This allows to run the program also from inner repository folders

The problem is in the call to `Repo` constructor which should get `'.'` for the current folder and not `'.git'`.

The current version of the code, from within other functions of this module, never checks `repo` was actually initialized. This lead to a number of other problems for which we'd probably like to show a nice error message too.
See for example this one:

```
:~/tmp$ semantic-release version --noop
Creating new version..
Traceback (most recent call last):
  File "/home/luca/.local/bin/semantic-release", line 11, in <module>
    load_entry_point('python-semantic-release', 'console_scripts', 'semantic-release')()
  File "/home/luca/.local/lib/python3.4/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/home/luca/.local/lib/python3.4/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/home/luca/.local/lib/python3.4/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/luca/.local/lib/python3.4/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/luca/.local/lib/python3.4/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/home/luca/Projects/python-semantic-release/semantic_release/cli.py", line 208, in cmd_version
    return version(**kwargs)
  File "/home/luca/Projects/python-semantic-release/semantic_release/cli.py", line 48, in version
    current_version = get_current_version()
  File "/home/luca/Projects/python-semantic-release/semantic_release/history/__init__.py", line 41, in get_current_version
    return get_current_version_by_tag()
  File "/home/luca/Projects/python-semantic-release/semantic_release/history/__init__.py", line 20, in get_current_version_by_tag
    version = get_last_version()
  File "/home/luca/Projects/python-semantic-release/semantic_release/vcs_helpers.py", line 39, in get_last_version
    for i in sorted(repo.tags, reverse=True, key=version_finder):
AttributeError: 'NoneType' object has no attribute 'tags'
```
This is due to try accessing `repo.tags` when `repo is None`.
To solve this other problem, I've added a sanity check in every function using `repo`: in case it is not correctly initialized it throws a `GitError` with a nice messge.

To complete the nice error reporting through the CLI this requires some actual catching, for which I added a single example [here](https://github.com/splendido/python-semantic-release/blob/splendido-fix-repo-init/semantic_release/cli.py#L49-L53) so that the above error now looks like this:

```
~/tmp$ semantic-release version --noop
Creating new version..
Not in a valid git repository
```

Please feel free to add more checks like the above one where you think they fit.

also, but minor, I've made a bit of linting to the module code.

### DIFF
--- a/semantic_release/vcs_helpers.py
+++ b/semantic_release/vcs_helpers.py
@@ -1,14 +1,19 @@
 import re
 
-from git import GitCommandError, NoSuchPathError, Repo, TagObject
+from git import GitCommandError, InvalidGitRepositoryError, Repo, TagObject
 
 from .errors import GitError
 from .settings import config
 
 try:
-    repo = Repo('.git', search_parent_directories=True)
-except NoSuchPathError:
+    repo = Repo('.', search_parent_directories=True)
+except InvalidGitRepositoryError:
     repo = None
+
+
+def check_repo():
+    if not repo:
+        raise GitError("Not in a valid git repository")
 
 
 def get_commit_log(from_rev=None):
@@ -16,6 +21,7 @@ def get_commit_log(from_rev=None):
     Yields all commit messages from last to first.
     """
 
+    check_repo()
     rev = None
     if from_rev:
         rev = '...{from_rev}'.format(from_rev=from_rev)
@@ -25,25 +31,34 @@ def get_commit_log(from_rev=None):
 
 def get_last_version(skip_tags=None):
     """
-    return last version from repo tags
+    Return last version from repo tags.
 
-    :return: a string contains version number
+    :return: A string contains version number.
     """
+
+    check_repo()
     skip_tags = skip_tags or []
 
-    def version_finder(x):
-        if isinstance(x.commit, TagObject):
-            return x.tag.tagged_date
-        return x.commit.committed_date
+    def version_finder(tag):
+        if isinstance(tag.commit, TagObject):
+            return tag.tag.tagged_date
+        return tag.commit.committed_date
 
     for i in sorted(repo.tags, reverse=True, key=version_finder):
-        if re.match('v\d+\.\d+\.\d+', i.name):
+        if re.match(r'v\d+\.\d+\.\d+', i.name):
             if i.name in skip_tags:
                 continue
             return i.name[1:]
 
 
 def get_version_from_tag(tag_name):
+    """
+    Gets commit hexsha for given ref tag.
+
+    :param tag_name: The name of the sought rference tag.
+    """
+
+    check_repo()
     for i in repo.tags:
         if i.name == tag_name:
             return i.commit.hexsha
@@ -53,9 +68,10 @@ def get_repository_owner_and_name():
     """
     Checks the origin remote to get the owner and name of the remote repository.
 
-    :return: a tuple of the owner and name
+    :return: A tuple of the owner and name.
     """
 
+    check_repo()
     url = repo.remote('origin').url
     parts = re.search(r'([^/:]+)/([^/]+).git$', url)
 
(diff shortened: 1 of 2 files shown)


semantic_release/vcs_helpers.py
```diff
@@ -25,25 +31,34 @@ def get_commit_log(from_rev=None):
...
     """
+
+    _check_repo()
     skip_tags = skip_tags or []
 
-    def version_finder(x):
-        if isinstance(x.commit, TagObject):
-            return x.tag.tagged_date
-        return x.commit.committed_date
+    def _version_finder(tag):
```
relekang: Why does this function name start with an `_`?

semantic_release/vcs_helpers.py
```diff
@@ -1,21 +1,27 @@
...
 
 try:
-    repo = Repo('.git', search_parent_directories=True)
-except NoSuchPathError:
+    repo = Repo('.', search_parent_directories=True)
+except InvalidGitRepositoryError:
     repo = None
 
 
+def _check_repo():
```
relekang: Why does this function name start with an `_`?

splendido: It is intended to be a private function, not parte of the public API
https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

For private functions no docstring is required...

semantic_release/vcs_helpers.py
splendido: It is intended to be a private function, not part of the public API
https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

For private functions no docstring is required...

relekang: Please remove the `_`.

semantic_release/vcs_helpers.py
relekang: Please remove the `_`.

semantic_release/vcs_helpers.py
```diff
@@ -1,21 +1,27 @@
...
+from git import GitCommandError, InvalidGitRepositoryError, Repo, TagObject
 
 from .errors import GitError
 from .settings import config
 
 try:
-    repo = Repo('.git', search_parent_directories=True)
-except NoSuchPathError:
+    repo = Repo('.', search_parent_directories=True)
+except InvalidGitRepositoryError:
```
relekang: Shouldn't `NoSuchPathError`still be here?

splendido: No, it shouldn't!
I had to change it in order to catch the exception raised by `Repo` in case provided path doesn't lead to a git reporitory: apparently the exception raised is `InvalidGitRepositoryError`.
`NoSuchPathError` is raised only in case you provide an invalid path, which is unlikely for the `'.'` representing the current folder: [see docs](https://gitpython.readthedocs.io/en/stable/reference.html?highlight=Repo#git.exc.NoSuchPathError).
With your version of the code is very likely you get `NoSuchPathError` when there's not `.git` folder from where you ran your command...

semantic_release/vcs_helpers.py
splendido: As you wish... done!

semantic_release/vcs_helpers.py
splendido: As you wish... done!

relekang:
Thanks :v:

PR #93, opened 2018-11-19
Change Dependancies on Requests

DavidWylie:
The existing hard dependancy on a version of the requests library limits how this can be used.
such that for some projects it cannot be used in the same container as testing or building the application.

The changes attached should make this a bit more flexible and allow versions of requests to be used which python-semantic-release is not using yet.


DavidWylie:
thanks for quick merge.

relekang:
This is released in v4.0.0

PR #94, opened 2018-11-20
feat: GitLab CI support and custom commit message

cvockrodt:
Custom commit message was really intended for use with GitLab CI to prevent pipelines from triggering, but will likely be useful elsewhere. I wrote this in as a breaking change...

GitLab CI support uses the following CI environment variables:
`CI_COMMIT_REF_NAME` since GitLab 9.0
`GITLAB_CI` since all
see [GitLab CI](https://docs.gitlab.com/ee/ci/variables/)


relekang:
🙌

relekang:
This is released in v4.0.0

PR #95, opened 2018-11-21
[WIP] docs: Add type hints and more complete docstrings

### LINKED ISSUE #81: Add type hints
Would accept a patch to add type hints throughout the project? I think they add self documentation as well as help catch bugs.

cvockrodt:
TODOs are included where return types are unknown. Includes a few style
changes suggested by pylint

Closes #81


cvockrodt:
python 2 tests fail as these type hints are for python 3...

semantic_release/__init__.py
```diff
@@ -1,11 +1,13 @@
+"""Semantic Release
```
relekang: Why was these added?

relekang:
Is there anyway we can strip these away for python2?

semantic_release/__init__.py
cvockrodt: I used pylint to help me catch all of the missing type docs and missing module docstrings showed up as well. I just threw those in there to silence those

cvockrodt:
mmm... I'm not really sure. You could certainly maintain two separate branches but that's a pain. I didn't realize until i'd already done all the work. IMO [python 2 is dying](https://pythonclock.org/) and you should think about phasing out support for it.

PR #102, opened 2019-01-10
fix(parser_angular): Fix non-match when special chars in scope

timbedard:
Per https://github.com/relekang/python-semantic-release/issues/101, this parser fails to match when the scope contains special characters such as: `.`, `,`, `*`, `&`, etc.

This fix opens up the contents of the scope to be more broad and supplies the corresponding tests.

(It also reuses the keys from `TYPES` to be a little cleaner and easier to maintain.)


relekang:
Thanks :v:

timbedard:
Happy to help. 👍

PR #103, opened 2019-01-17
Maintain version variable formatting on bump

mcmcgrath13:
Small change to the way the version is written to the config file it is read from.  This allows the formatting to be the same as before semantic-release changed it.

Prior behavior
`my_version_var="1.2.3"` => `my_version_var = '1.2.4'`

New behavior
`my_version_var="1.2.3"` => `my_version_var="1.2.4"`

I am using python-semantic-release with a Julia project and this change will allow for consistent formatting in my Project.toml file where the version is maintained

### DIFF
--- a/semantic_release/history/__init__.py
+++ b/semantic_release/history/__init__.py
@@ -106,11 +106,10 @@ def set_new_version(new_version: str) -> bool:
         content = fr.read()
 
     content = re.sub(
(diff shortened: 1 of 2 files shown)


relekang:
Thanks👍
Do you mind adding a test for this?

mcmcgrath13:
all set

relekang:
Thanks :v:

PR #105, opened 2019-01-30
Add support for bitbucket CI

alexsavio:


semantic_release/ci_checks.py
```diff
@@ -106,3 +118,5 @@ def check(branch: str = 'master'):
         circle(branch)
     elif os.environ.get('GITLAB_CI') == 'true':
         gitlab(branch)
+    elif os.environ.get('CI') == 'true':
```
relekang: This should use something specific for bitbucket since most services set `CI=true`.

alexsavio: I see, there is no 'BITBUCKET_CI' though, it could be another kind of condition like: `'BITBUCKET_BUILD_NUMBER'  in os.environ` ? https://confluence.atlassian.com/bitbucket/environment-variables-in-bitbucket-pipelines-794502608.html

relekang: Sounds like a good idea 👍

alexsavio:
Regarding the flake8 comment. My confusion is that flake8 is not even a dev dependency from this module. Also in my Python projects, to make all the process uniform, I usually put a Makefile like this: https://github.com/alexsavio/hansel/blob/master/Makefile. So I can ask contributors to run their tests locally (through `make`) so they know what they might expect from the CI. What do you think if I put one of these (with the necessary adaptations), and add flake8 as a dependency?

semantic_release/ci_checks.py
alexsavio: Done

relekang:
Tox handles all of these things running `tox` in the project directory will do the same as CI does.

relekang:
This is released in v4.1.0

PR #110, opened 2019-02-08
Update documentation about commands and parsing.

kmpm:


docs/commands.rst
```diff
@@ -0,0 +1,36 @@
+.. _commands:
+
+Commands
+--------
+
+.. _cmd-changelog:
+
+changelog
+^^^^^^^^^
+When executed this command will generate/update ``CHANGELOG.txt``.
```
relekang: This is not true, the changelog command prints the changelog to stdout. If the option `--post` is passed it will be posted on github using the `GH_TOKEN` similar to what publish does.

docs/commands.rst
```diff
@@ -0,0 +1,36 @@
...
+#. If the environment variable ``GH_TOKEN`` (or equivalent for your
+   vcs provider) is set then :ref:`cmd-changelog` will be executed and
+   the changelog will be posted to your vcs provider if that is supported.
+
+
+.. _cmd-version:
+
+version
+^^^^^^^
+This will only figure out the new version number using
```
relekang: I think it would make sense to remove only from this sentence.

relekang:
Thanks 😄

Issue #38, opened 2015-12-21
Start logging


relekang:
There should be a lot more debug logging than there is now, in order to find out why something is not working.

kmpm:
What method would be prefered to use?
Python logging or something like https://pypi.org/project/pydebug?

relekang:
Looks perfect 👌

PR #113, opened 2019-02-11
refactor: add debug output

kmpm:
Possible fix for #38 using https://pypi.org/project/ndebug/

To test output
```bash
DEBUG=* semantic-release changelog
```


kmpm:
Documentation... sure. I'll add some.
And you could perhaps give some hints about where to add more debug output. Any special area that users have troubble with?

relekang:
Oh, that was a bunch of documentation 😄 Thank you so much 🙌 

> And you could perhaps give some hints about where to add more debug output. Any special area that users have trouble with?

I think this is a good start, cannot think of something on the top of my head that is not covered by this pr.

relekang:
This was just released as 4.1.1 🎉

Issue #129, opened 2019-06-11
Flake8 does not pass on `tox -e flake8`


esciara:
Ran flake8 on master today and got following issues:
```
$ tox -e flake8
flake8 installed: entrypoints==0.3,flake8==3.7.7,mccabe==0.6.1,pycodestyle==2.5.0,pyflakes==2.1.1,python-semantic-release==4.1.1
flake8 run-test-pre: PYTHONHASHSEED='3454641464'
flake8 run-test: commands[0] | flake8
./semantic_release/__init__.py:7:1: F401 '.errors.ImproperConfigurationError' imported but unused
./semantic_release/__init__.py:7:1: F401 '.errors.SemanticReleaseBaseError' imported but unused
```

relekang:
Master is green now ✌️

PR #130, opened 2019-06-11
Dev cleanup: fixing various issues

esciara:


esciara:
Beware of false negative on tests: see #131.

semantic_release/__init__.py
```diff
@@ -3,10 +3,6 @@
 __version__ = '4.1.1'
 
 
-from .errors import UnknownCommitMessageStyleError  # noqa; noqa
```
relekang: Why are these removed? This is a breaking change, since this will break imports of errors directly from `semantic_release`. Can you please add them back?

esciara: Yes, I removed them as these caused flake8 errors (see #129), and tests are passing when removing these imports. So it is not clear what pb this causes.

Can you clarify the pb with an actual test or specific example?

esciara: hi @relekang ,

A kind reminder. Could you clarify please? Would love to work on having this PR merged.

relekang: Anyone who has installed this package can have done `from semantic_release import UnknownCommitMessageStyleError` and next release after this will break this so thus we need a major release in order to do it.

PR #141, opened 2019-08-29
fix: manage subgroups in git remote url

### LINKED ISSUE #139: Tests fail when owner is not relekang
```
________ test_get_repository_owner_and_name ________

    def test_get_repository_owner_and_name():
>       assert get_repository_owner_and_name()[0] == 'relekang'
E       AssertionError: assert 'SewanDevs' == 'relekang'
E         - SewanDevs
E         + relekang

tests/test_vcs_helpers.py:52: AssertionError
```

To work on this project I forked it under SewanDevs (for 119) and the test breaks... Maybe we could just test that it is an alphanumeric string ? Or maybe mock.

### LINKED ISSUE #140: Manage subgroups in origin url
Today, the method `get_repository_owner_and_name` returns owner and name as the last two parts of the origin url. It can then use it to form the url to reach the repository.

The problem in the implementation is that it doesn't support subgroups like in gitlab (and I think in bitbucket). Apparently, a github origin url always follow the format : `git@github.com:<owner>/<name>.git`. But in gitlab we can get something like `git@gitlab.example.com:<group>/<subgroup>/<name>.git`. The current method here would return `name` correctly but only `subgroup` as owner. That would not allow the use of those as reference url.

I suggest to change the regular expression used in this method to get `<group>/<subgroup>` as owner instead of just `<subgroup>`. I can start working on this in the branch I created for #121

fattybenji:
This is a necessary fix for gitlab integration.
For an illustration of the need and use for this fix, test was edited.

Fixes #139
Fixes #140

### DIFF
--- a/semantic_release/vcs_helpers.py
+++ b/semantic_release/vcs_helpers.py
@@ -85,7 +85,7 @@ def get_repository_owner_and_name() -> Tuple[str, str]:
 
     check_repo()
     url = repo.remote('origin').url
-    parts = re.search(r'([^/:]+)/([^/]+).git$', url)
+    parts = re.search(r'[:/]([^\.:]+)/([^/]+).git$', url)
     if not parts:
         raise HvcsRepoParseError
     debug('get_repository_owner_and_name', parts)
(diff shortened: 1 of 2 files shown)


relekang:
Thanks

PR #143, opened 2019-09-03
feat: allow users to get version from tag and write/commit bump to file

### LINKED ISSUE #104: python-semantic-release should have an option to pull version_variable from git tag
python-semantic-release should have an option to pull version_variable from git tag

python-semantic-release requires version_variable to be in a file and bumps and auto-commits it.

For parity with semantic-release per https://semantic-release.gitbook.io/semantic-release/support/faq#why-is-the-package-jsons-version-not-updated-in-my-repository semantic release updates package.json then prepares and publishes the package and doesn't auto-commit the version change. 

This is desirable for CI so that auto-commits don't require troublesome exceptions to prevent an infinite loop of version bumps.

If we don’t push the commit to the version_variable subsequent runs report the following error
```$ semantic-release version --patch
Creating new version..
Current version: 0.0.0
Cmd('git') failed due to: exit code(128)
  cmdline: git tag -m v0.0.1 -a v0.0.1
  stderr: 'fatal: tag 'v0.0.1' already exists'
```

fattybenji:
Before this commit, version was bumped in the file, but only committed
if version was obtained from `version_variable` (version_source == `commit`).
Also added a relevant test and a description for this new option.

Fixes #104

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -96,7 +96,9 @@ def version(**kwargs):
         return True
 
     set_new_version(new_version)
-    if config.get('semantic_release', 'version_source') == 'commit':
+    if config.get('semantic_release', 'commit_version_number', fallback=(
+                config.get('semantic_release', 'version_source') == 'commit')
+            ):
         commit_new_version(new_version)
     tag_new_version(new_version)
     click.echo('Bumping with a {0} version to {1}.'.format(level_bump, new_version))
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -53,3 +53,7 @@ Moreover, those configuration values can be overloaded with the ``-D`` option, l
 ``hvcs``
     The name of your hvcs. Currently only ``github`` and ``gitlab`` are supported.
     Default: ``github``
+
+``commit_version_number``
+    Whether or not to commit changes when bumping version.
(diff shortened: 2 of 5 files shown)


semantic_release/cli.py
```diff
@@ -86,7 +86,9 @@ def version(**kwargs):
         return True
 
     set_new_version(new_version)
-    if config.get('semantic_release', 'version_source') == 'commit':
+    if config.get('semantic_release', 'commit_changes', fallback=(
```
relekang: Will commit_changes set to false and version_source set to commit make it commit the changes? Also maybe commit_version_number is a better name?

fattybenji: In that case, it would not commit the change. Setting `commit_changes` overloads other variables. I did it that way in order to keep the current behavior for everyone not interested in changing it. This way, in your example, the source is still defined by `version_source`, but it does not mean that the changes to the version file are commited. I don't know why someone would want to do that, but I think that if we allow people to get from vcs and write in file, we should also allow them to get from the file and not write in this file.

I will change the name to `commit_version_number` if you think it's clearer.

I will also include the change I asked you about in history.logs.evaluate_version_bump at line 44.

relekang: Sounds good 👍

relekang:
Can you rebase this?

fattybenji:
Rebased

PR #146, opened 2019-09-03
fix: update list of commit types to include build, ci and perf

### LINKED ISSUE #145: Update commit types
The angular list contains the following types that are not recognized by this library :

* build
* ci
* perf

See here for the reference for my issue : https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines

The `chore` type is not included in the spec, but it is included in this library (however it is useful to have no impact on the version).

Moreover, the `perf` type should be added to `patch`-triggering types (see https://github.com/semantic-release/semantic-release/blob/master/docs/support/FAQ.md#how-can-i-change-the-type-of-commits-that-trigger-a-release)

fattybenji:
Also added perf to the types that trigger a patch update

Fixes #145


relekang:
Thanks 👍

PR #147, opened 2019-09-10
test: better test coverage

fattybenji:
Adds some coverage mainly on cli and vcs_helpers

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,9 +1,11 @@
 from click.testing import CliRunner
 
 import semantic_release
-from semantic_release.cli import main
+from semantic_release.cli import changelog, main, publish, version
+from semantic_release.errors import GitError, ImproperConfigurationError
 
 from . import mock, pytest, reset_config
+from .mocks import mock_version_file
 
 assert reset_config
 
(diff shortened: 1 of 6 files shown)


tests/test_vcs_helpers.py
```diff
@@ -14,6 +16,12 @@ def mock_git(mocker):
     return mocker.patch('semantic_release.vcs_helpers.repo.git')
 
 
+@mock.patch('semantic_release.vcs_helpers.repo', None)
+def test_check_repo():
```
relekang: This test name should reflect what it checks.

tests/test_pypi.py
```diff
@@ -39,3 +40,7 @@ def test_upload_with_custom_path(self, mock_run):
         self.assertEqual(args[2], mock.call(
             'twine upload -u username -p password  custom-dist/*'),)
         self.assertEqual(args[3], mock.call('rm -rf custom-dist'))
+
+    def test_raises_error(self):
```
relekang: This test name should reflect what it checks, something like `test_raises_error_when_missing_credentials`.

tests/test_cli.py
```diff
@@ -341,9 +473,81 @@ def test_changelog_should_call_functions(mocker, runner):
...
+
+
+def test_changelog_raises_exception(mocker):
+    mocker.patch('semantic_release.cli.get_current_version', return_value=None)
+    with pytest.raises(ImproperConfigurationError):
+        changelog()
+
+
+@mock.patch('semantic_release.cli.debug')
+def test_main_debug(mock_debug, runner):
```
relekang: What is the value in this test?

tests/test_cli.py
```diff
@@ -341,9 +473,81 @@ def test_changelog_should_call_functions(mocker, runner):
...
+    mock_previous_version.assert_called_once_with('current')
+    mock_generate_changelog.assert_called_once_with('current', None)
+    mock_markdown_changelog.assert_any_call('current', 'super changelog', header=False)
+    mock_check_token.assert_called_once_with()
+    mock_get_owner_name.assert_called_once_with()
+    mock_post_changelog.assert_called_once_with('owner', 'name', 'current',
+                                                'super md changelog')
+
+
+def test_changelog_raises_exception(mocker):
```
relekang: This test name should reflect what it checks, as in answer why do we expect an error here?

tests/test_cli.py
fattybenji: hum... I think that's a work in progress that stayed there, but it actually seems to hit the lines under main when debug is enabled. I can remove it if you like.

relekang: Let's just leave it.

PR #148, opened 2019-09-16
fix: support repo urls without git terminator

iamogbz:
Addresses https://github.com/relekang/python-semantic-release/issues/109#issuecomment-531053386


relekang:
Thanks 👍

PR #152, opened 2019-10-03
fix: update regex to get repository owner and name for project with dots

### LINKED ISSUE #151: Fix regex for repository names containing a point
It looks like the point added in the second capture group can break the regex if a project name contains a point. I think we want to support this.

It was introduced in the following commit : https://github.com/relekang/python-semantic-release/commit/700e9f18dafde1833f482272a72bb80b54d56bb3

However, just removing the point causes a problem, where the `.git` part is included in the repository name.

fattybenji:
Remove the dot from the second capture group to allow project names
containing dots to be matched.
Instead of a greedy '+' operator, use '*?' to allow the second group to
give back the '.git' (to avoid including it in the project name)

Fixes #151

### DIFF
--- a/semantic_release/vcs_helpers.py
+++ b/semantic_release/vcs_helpers.py
@@ -85,7 +85,7 @@ def get_repository_owner_and_name() -> Tuple[str, str]:
 
     check_repo()
     url = repo.remote('origin').url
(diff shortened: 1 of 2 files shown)


relekang:
Thanks, it is released in v4.3.2

PR #157, opened 2019-12-13
fix: fallback to whole log if correct tag is not available

### LINKED ISSUE #51: Fallback to whole log if correct tag is not available
The log evaluation looks at `v{current_version}...` in some cases(e.g. on a unreleased project). This will not be available. It should fall back to the whole log if that happens and the message generator should stop when it hit a commit containing the current version like it worked before.

fattybenji:
The method getting all commits to consider for the release will now test
whether the version in input is a valid reference. If it is not, it will
consider the whole log for the repository.

evaluate_version_bump will still consider a message starting with the
version number as a breaking condition to stop analyzing.

Fixes #51

I could not find a simple way to test that.

### DIFF
--- a/semantic_release/vcs_helpers.py
+++ b/semantic_release/vcs_helpers.py
@@ -7,6 +7,7 @@ from typing import Optional, Tuple
 
 import ndebug
 from git import GitCommandError, InvalidGitRepositoryError, Repo, TagObject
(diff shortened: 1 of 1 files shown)


relekang:
Thanks 🙌

PR #160, opened 2020-01-03
Implement exclamation points for breaking changes

### LINKED ISSUE #156: Implement the use of exclamation points to describe breaking changes
According to https://www.conventionalcommits.org/en/v1.0.0/ , a breaking change can be described by using an exclamation point. Here are three cases that will generate the same major version bump and the same breaking change message.

Case 1 :
```
feat!: completely change this behavior
```

Case 2 :
```
feat(versioning)!: completely change this behavior
```

Case 3 :
```
feat!: update the parser for commits

BREAKING CHANGE: completely change this behavior
```

The exclamation point is optional, and if used, is just before the colon.

If there is an exclamation point, the `BREAKING CHANGE` section is optional, and the commit message will be used as breaking change description.

fattybenji:
This PR is mostly for #156 but I also added a change for point 16 of the specification : https://www.conventionalcommits.org/en/v1.0.0/

* Implement the use of exclamation points to describe breaking changes
* Allow the use of `BREAKING-CHANGE` as synonym of `BREAKING CHANGE`

I can put the second commit in another branch/PR if you prefer.

Fixes #156

### DIFF
--- a/semantic_release/history/logs.py
+++ b/semantic_release/history/logs.py
@@ -110,15 +110,22 @@ def generate_changelog(from_version: str, to_version: str = None) -> dict:
 
             changes[message[1]].append((_hash, message[3][0]))
 
-            if message[3][1] and 'BREAKING CHANGE' in message[3][1]:
-                parts = re_breaking.match(message[3][1])
-                if parts:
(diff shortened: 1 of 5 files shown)


relekang:
Thanks 👍

Issue #109, opened 2019-02-08
Add support for Github actions


relekang:
- [x] Add necessary checks for github actions.
- [x] Setup github actions for releasing this project by calling on the current source code.
- [x] Add to the documentation how to setup semantic release with github actions.

relekang:
It would be awesome if you have the time to create a PR for it 😄

iamogbz:
@relekang also I had temporarily fixed this in my CI by appending `.git` to the origin but had some other issues that prevented me from using the package e.g. `--noop` still pushing tags etc. I will create issues as I don't think those are github actions specific.

alandtse:
I'm testing the merged version and using `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` which uses the default GITHUB_TOKEN. This results in the following error while trying to push:
 
```
stderr: 'fatal: could not read Password for 'https://[AUTH_TOKEN]@github.com': No such device or address'
```
I believe the problem is the default `GITHUB_TOKEN` expects to be used with a `GITHUB_ACTOR` as the username. I'm basing this observation on this [code](https://github.com/mikeal/publish-to-github-action/blob/master/entrypoint.sh#L10). 

I'll play around and see if it's as simple as adding `GITHUB_ACTOR` as the user in the https string.

alandtse:
I ended up getting farther with my [repo changes](https://github.com/relekang/python-semantic-release/compare/master...alandtse:master). However, ran into a branch protection issue. 
```
stderr: 'remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: You're not authorized to push to this branch. Visit https://help.github.com/articles/about-protected-branches/ for more information.        
```
I wasn't sure if the problem was because I'm committing to an organization, so I gave up with the `GITHUB_TOKEN` _(I'm an admin on the repo I'm pushing to so not sure why the `GITHUB_TOKEN` isn't enough to log in as my account and didn't want to debug it further)_ and created a personal access token.  With that I was able to commit using my changes.
  
**However**, I don't know if my changes are necessary with a personal access token so I'm leaving this comment in case someone wants to try the code with a `GITHUB_TOKEN`.  If it works, happy to do a PR or you can just integrate my code directly.

AlphaMycelium:
@alandtse I've tried your changes with a `GITHUB_TOKEN`, and can confirm that it works correctly.

alandtse:
@AlphaMycelium Great, I'll put in a PR.

[...]

PR #162, opened 2020-01-16
fix(github): add check for GITHUB_ACTOR for git push

alandtse:
Partially addresses #109; tested by @AlphaMycelium


relekang:
Thanks 👍

PR #172, opened 2020-02-17
Refactor GitHub API calls into separate functions

AlphaMycelium:
The following changes have been made to the `Github` class in `hvcs.py`:

- Separate each type of request into its own function
    - Function docstrings include links to the GitHub API documentation
- Rename the variable `status` to `success`, as the word "status" may imply that it contains the numerical status code of the request
- Make debug logging more easily understandable

This should not affect the actual behaviour of the code.

### DIFF
--- a/semantic_release/hvcs.py
+++ b/semantic_release/hvcs.py
@@ -86,9 +86,33 @@ class Github(Base):
 
         :return: The status of the request
         """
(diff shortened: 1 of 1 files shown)


AlphaMycelium:
I'm going to re-do this as part of #177, as when that is merged it will cause a merge conflict and additional changes will be required anyway.

PR #155, opened 2019-11-05
Lower the dependency on click to 2.0.

openbrian:
Not sure what depends on click < 6.0, but something in python-semantic-release does.  Naturally this conflicts with ~=7.0.

I ran tests and they pass.

See #117 

Fixes #154

PR #173, opened 2020-02-17
fix: Bump dependencies and fix Windows issues on Development

### LINKED ISSUE #164: Update invoke version to latest
the current invoke version specified in the requirements file is set to 0.11.1. Theres a maximum recursion error that occurs with this version i manually updated to version 1.4.1 and the recursion error is gone. Its related to this issue https://github.com/pyinvoke/invoke/issues/257.

KingDarBoja:
Fixes #164

The tests were not passing on Windows development mode due to path separator issues but made some fixes to allow that. Still 2 left missing to fix.

Also, bump required dependencies version and add extra classifiers. So far, it does work on Python 3.8 too.


requirements/base.txt
```diff
@@ -1,7 +1,7 @@
-click>=2.0, <8.0
-gitpython~=2.1
-invoke==0.11.1
-semver~=2.2
+click>=7.0, <8.0
```
relekang: We have tried to update click before but having this tight a range on click causes problems for users and we do not use any versions specific to seven so please put this back?

tests/test_settings.py
```diff
@@ -35,7 +39,8 @@ def test_defaults(self, mock_getcwd):
     @mock.patch('semantic_release.settings.getcwd', return_value='/tmp/')
     def test_toml_override(self, mock_getcwd):
         # create temporary toml config file
-        dummy_conf_path = '/tmp/pyproject.toml'
+        dummy_conf_path = os.path.join(here, 'tmp', 'pyproject.toml')
```
AlphaMycelium: This change is causing `True` to be returned for `upload_to_pypi` when it should be `False`.

tests/test_settings.py
```diff
@@ -63,12 +68,14 @@ def test_toml_override(self, mock_getcwd):
     @mock.patch('semantic_release.settings.getcwd', return_value='/tmp/')
     def test_no_raise_toml_error(self, mock_getcwd, mock_debug):
         # create temporary toml config file
-        dummy_conf_path = '/tmp/pyproject.toml'
+        dummy_conf_path = os.path.join(here, 'tmp', 'pyproject.toml')
```
AlphaMycelium: This change is causing the test to fail.

tests/test_settings.py
KingDarBoja: The problem is that on Windows, the path isn't being recognized if it contains `/`. The file does exist and has the proper content as stated on the next step but for some reason it isn't properly loaded on the assertion.

tests/test_settings.py
KingDarBoja: Same comment as above.

requirements/base.txt
KingDarBoja: It should be covered on the test environment, as it should work on python 3.4 and higher.

I don't have any problem lowering `click` to 2.0 as minimum required version but should be good to know which issues are caused by bumping this dependency.

tests/test_settings.py
AlphaMycelium: Do these changes make the test pass on Windows? If so, we could add an OS-based switch using something like `platform.system()`.

requirements/base.txt
relekang: See #155 for more info

tests/test_settings.py
```diff
@@ -12,6 +12,10 @@
 assert reset_config
 
 
+# Set path to this directory
+here = os.path.abspath(os.path.dirname(__file__))
```
relekang: Maybe change this to 
```suggestion
temp_dir = os.path.abspath(os.path.dirname(__file__)) if platform.system() == "Windows" else "/tmp/"
```
and use it below?

tests/test_settings.py
KingDarBoja: These two don't pass but due to the `AssertionError`. Before this change, Windows wasn't able to find the file path as the path separator had to be `\\` instead of `/`. Hence, throwing an error at the `pathlib` script, which is inside of pytest package.

tests/test_settings.py
KingDarBoja: I will test it again as soon as possible 👍

requirements/base.txt
KingDarBoja: I see, so far all test are passing with `click>= 7.0`. I will check again the version and share the test results.

tests/test_settings.py
AlphaMycelium: Aaah sorry, that's what's causing it on linux too. In that case we need to find what else is causing the `AssertionError`s.

KingDarBoja:
After testing the new changes, the two tests are still failing with same AssertionErrors:
- **ConfigTests.test_no_raise_toml_error:** `AssertionError: Expected 'debug' to be called once. Called 0 times.`
- **ConfigTests.test_toml_override:** `AssertionError: True != False`

Also, getting this error if using `tox` on my conda env (It uses `tox-conda` and has been working before with other projects`).

```console
OSError: [WinError 123] El nombre de archivo, el nombre de directorio o la sintaxis de la etiqueta del volumen no son correctos: 'D:\\Programming\\py_projects\\python-semantic-release:D:\\Programming\\py_projects\\python-semantic-release'
Coverage.py warning: No data was collected. (no-data-collected)
ERROR: InvocationError for command 'D:\Programming\py_projects\python-semantic-release\.tox\py3-tests\Scripts\coverage.EXE' run -p --source=semantic_release -m py.test -v tests (exited with code 1)
```

But still the coverage is being retrieved:

image

tests/test_settings.py
KingDarBoja: `config.getboolean('semantic_release', 'upload_to_pypi')` is returned as `True` but the `pyproject.toml` file has it as false value. 🤔

KingDarBoja:
I think I found the issue, atleast for **ConfigTests.test_no_raise_toml_error**: Looks like the `current_dir` value on the `mock_debug` call isn't being properly set as the `mock_getcwd` does match the correct path.

Below is the print call on the `mock_debug` call for the `current_dir` and `toml_conf_path` variables.
```console
Current dir
D:\Programming\py_projects\python-semantic-release
TOML Path
D:\Programming\py_projects\python-semantic-release\pyproject.toml
```

KingDarBoja:
I managed to setup and test this package with multiple python versions (from 3.4 to 3.8) and the only two currently failing are `3.4` and `3.5`. Looks like somewhere on the `pypi.py` file (and probably on others) are relying on f-strings, which was introduced on python 3.6+. 

@relekang Do you want me to submit this config of the `tox.ini` file so it can help you track the python support?

AlphaMycelium:
@KingDarBoja Hmmm, either we'll have to remove the f-strings, or stop supporting those versions of Python.

KingDarBoja:
@AlphaMycelium This is the entire log after running the tox command with the new setup, using conda and `tox-conda`:

[entire_log_tox_pysemver.txt](https://github.com/relekang/python-semantic-release/files/4231168/entire_log_tox_pysemver.txt)

You will find all the errors and detailed info can be found inside other logs but this should be enough to catch them

AlphaMycelium:
@KingDarBoja That's definitely being caused by the use of f-strings and type annotations in versions before Python 3.6.

setup.py
```diff
@@ -48,5 +48,9 @@ def _read_long_description():
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.4',
```
relekang: The current version does not work with 3.4 and 3.5 so let's remove them.

requirements/base.txt
relekang: Okay, let's leave it and see how it goes.

KingDarBoja:
@AlphaMycelium I accidentally added py38 to the tox env but for some reason, the CircleCI is failing to find the python interpreter, which is weird as it is supported by [CircleCI](https://circleci.com/docs/2.0/circleci-images/#python).

KingDarBoja:
Well after a quick test, look like CircleCI has issues with tox + python environment ranges.

Setting the base python as python3.8 will install py38 and py37 from the tox env but fail on py36. If setting the base python as python3.6, will install py36 and py37 from the tox env but fail on py38. 🤔

KingDarBoja:
@relekang Done!

PR #183, opened 2020-02-28
feat(git): Add a new config for commit author

alissonperez:
I realized that it commits with a specific author `semantic-release <semantic-release>`. I added a new config to make it possible to be changed.  Please, take a look guys!

Thanks!


relekang:
Thanks 👍

PR #184, opened 2020-03-01
feat: make commit message configurable

claymcleod:
On my project, we use Conventional Commit messages: https://www.conventionalcommits.org/en/v1.0.0/. This requires the commit message to be more configurable.


semantic_release/vcs_helpers.py
```diff
@@ -120,8 +120,13 @@ def commit_new_version(version: str):
     """
 
     check_repo()
+    commit_subject = config.get('semantic_release', 'commit_subject')
     commit_message = config.get('semantic_release', 'commit_message')
-    message = '{0}\n\n{1}'.format(version, commit_message)
+
+    message = commit_subject.format(version=version)
+    if commit_message:
```
claymcleod: There is one case which is not _exactly_ the same here: any user which previously set the `commit_message` as `""` would technically get a commit of

```
{version}

```

Here, we check to see if the commit message is empty: if it is not, only then would the two newlines be added. So this is the real difference, it's not technically backwards compatible but I think it is an improvement. If you feel you don't want to include it until 5.0, you could just remove the `if` statement.
```

relekang:
Thanks, I am thinking we should change the default values to be compatible with conventional commits. Do you know what we need to change?

claymcleod:
After doing some research, it appears there is no strong consensus for how to treat this due to (1) differing styles being using and (2) it's just not explicitly laid out in the spec. Look at [the project which use conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#projects-using-conventional-commits) for inspiration:

* [Angular](https://github.com/angular/angular/) uses the style `release: {version}`
* [Commitlint](https://github.com/conventional-changelog/commitlint) uses `{version}`
* [Electron](https://github.com/electron/electron) uses the style `Bump {version}`
* [yargs](https://github.com/yargs/yargs) uses the style `chore: release {version}`

So I think it's safe to use whatever you want. For my project, we like having the same commit format for all commits, so we settled on `chore: bump version to {version}` to ensure compatibility out of the box with the (angular style guide)[https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular]

claymcleod:
To conclude, I think it's fine to keep your current default as long as the message is configurable, but I'm happy to change it in my PR to whatever you think 😄

relekang:
Okay, let's keep the default the same for now.

claymcleod:
Great! Thank you!

PR #191, opened 2020-03-18
Set up releasing with GitHub Actions

AlphaMycelium:
This was listed as one of the tasks on #109:
> Setup github actions for releasing this project by calling on the current source code.

I have set up a workflow to run on pushes to the `master` branch. For this to work, **the `PYPI_PASSWORD` secret must be set [here](https://github.com/relekang/python-semantic-release/settings/secrets/) by @relekang before merging.**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,5 +16,5 @@ jobs:
       uses: AlphaMycelium/psr-action@v0.2.1
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
-        pypi_username: relekang
+        pypi_username: ${{ secrets.PYPI_USERNAME }}
         pypi_password: ${{ secrets.PYPI_PASSWORD }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,20 @@
+name: Semantic Release
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Python Semantic Release
+      uses: AlphaMycelium/psr-action@v0.2.1
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        pypi_username: relekang
```
relekang: Can we put this in secrets as well? I am planning to create a separate user for this.

relekang:
@AlphaMycelium is there any way we can make this wait on the build from circle ci marking master as ok?

AlphaMycelium:
@relekang Take a look at [wait-for-status](https://github.com/WyriHaximus/github-action-wait-for-status), although I've not used it before so I have no idea how or if it will work.

Alternatively, the CircleCI checks could be migrated to GitHub Actions. We would then create a workflow with one job for testing and a second for releases. The release job would only run on master. https://stackoverflow.com/a/58142412/6152171

PR #195, opened 2020-03-19
feat(build): allow config setting for build command

### LINKED ISSUE #188: Support poetry
Currently trying to use `python-semantic-release` along with `poetry` to build and release my project. 

Unfortunately, `semantic-release` is failing with the error message: 

```
Current version: 2.1.5
Creating new version..
Current version: 2.1.5
Bumping with a patch version to 2.1.6.
Pushing new version
Building distributions
python: can't open file 'setup.py': [Errno 2] No such file or directory
Command execution failure!

Exit code: 2

Stderr:

python: can't open file 'setup.py': [Errno 2] No such file or directory
```

What needs to be done to allow this setup to work? My project does not have a `setup.py`, only a `pyproject.toml`.

alandtse:
BREAKING CHANGE: The build_commands configuration variable has been renamed to build_command and changed to allow defining the entire command as opposed to parameters.
Closes #188


docs/configuration.rst
```diff
@@ -62,7 +62,7 @@ Moreover, those configuration values can be overloaded with the ``-D`` option, l
     Flag for whether the dist folder should be removed after a release. Default: true
 
 ``build_commands``
-    Parameters to call setup.py with to build dists. Default: ``sdist bdist_wheel``
+    Command to build dists. Default: ``python setup.py sdist bdist_wheel``
```
AlphaMycelium:
```suggestion
    Command to build dists. Build output should be stored in the directory configured in `dist_path`. Default: ``python setup.py sdist bdist_wheel``
```

semantic_release/dist.py
```diff
@@ -7,7 +7,7 @@
 
 def build_dists():
     commands = config.get('semantic_release', 'build_commands')
-    run(f'python setup.py {commands}')
+    run(f'{commands}')
```
relekang: Let's remove the extra string interpolation.

```suggestion
    run(commands)
```

semantic_release/dist.py
```diff
@@ -7,7 +7,7 @@
 
 def build_dists():
     commands = config.get('semantic_release', 'build_commands')
```
relekang: Can we change `build_commands` to `build_command` since we are doing a breaking change already?

```suggestion
    commands = config.get('semantic_release', 'build_command')
```

alandtse:
Just realized that there are two breaking changes in this PR and I only mentioned it in the first commit. Will you remember to update the merge commit to include the updated description I made in the first post?

AlphaMycelium:
Uhm... looks like we need to rename `build_commands` in defaults.cfg...
https://github.com/relekang/python-semantic-release/blob/149138205c900fb7ff1a56359dab66a3bf64a70c/semantic_release/defaults.cfg#L15

relekang:
Yeah, on it 👍

PR #197, opened 2020-03-19
Move GitHub Action into this repository

### LINKED ISSUE #196: Move GitHub Action into main repository
Do we think it is a good idea to move [psr-action](https://github.com/AlphaMycelium/psr-action) (Python Semantic Release for GitHub Actions) into a sub-directory within this repository, or should it be kept separate?

I am happy to make a pull request for this if we decide to go ahead with it. My separate repository would be archived and kept around to avoid breaking any existing setups.

## Pros

- More organised
- A specific version of Python Semantic Release can be loaded by tag or commit reference
- An update to the action can be released at the same time as any breaking changes

## Cons

- Versions prior to this change would not be available to use on Actions
- Anyone who has set up the action using my repository will need to update their configuration
- Cannot be published to GitHub Marketplace if we store it in a sub-directory

AlphaMycelium:
Move [psr-action](https://github.com/AlphaMycelium/psr-action) into this repository. Fixes #196.

Documentation for the action is in `action.rst`. We may wish to add a link to this somewhere in the docs, and possibly move the file. I chose to place the action metadata directly into the root of the repository as `Dockerfile` needs to be there anyway as there is currently no other way to change the build context.

This also means that it is possible to publish the action on the GitHub Marketplace if we want to do so in the future, although this is not at all necessary for it to function (some changes to `readme.rst` would be required before publishing).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,12 +8,13 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    if: github.repository == 'relekang/python-semantic-release'
 
     steps:
     - uses: actions/checkout@v1
 
     - name: Python Semantic Release
-      uses: AlphaMycelium/psr-action@v0.2.1
+      uses: ./
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


relekang:
Awesome 👍

PR #201, opened 2020-03-23
Use a named tuple for parsed commits

AlphaMycelium:
This improves readability as we can use attributes such as `bump` and `descriptions` instead of confusing numeric indices.


relekang:
Nice 👍

PR #202, opened 2020-03-24
Run release daily instead of on push

AlphaMycelium:
This gives a chance to merge multiple pull requests or create multiple commits without each one triggering an individual release.

For example, v5.0.1 and v5.0.2 were created on the same day with only one commit in each of them. Running daily would have created a single release with both commits.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,9 +1,9 @@
 name: Semantic Release
 
 on:
-  push:
-    branches:
-      - master
+  schedule:
+    # Release changes once a day
+    - cron: '0 0 * * *'
 
 jobs:
   release:


relekang:
I am not sure about this. I see a few benefits of it happening on merge:
* Things get released quicker.
* Someone that can fix things are awake, because it is related to something being merged.

Based on the history of the project, I think multiple releases in a day is not that common.

AlphaMycelium:
@relekang Not a problem, we'll leave it as it is.

Just be aware that multiple pushes within the space of a minute or two (while the action for the previous push is still running) could create a race condition.

relekang:
Yes, it is a way to block multiple releases since it will not be allowed to push 😅

PR #207, opened 2020-04-01
Changelog sections customization

### LINKED ISSUE #193: Changelog is incomplete
Hi, thanks for this project, in first place.

I want to start using this project but I'm having difficulties in generating a proper changelog.

We _just started_ using commitizen. So there's thousands of commits before the ones using commitizen. We have about 12 commits using commitizen. We have `build`, `fix`, `revert`, `feat`, `fix`, `refactor` and `improvement.

Perhaps I'm misunderstanding how to use it, but the changelog has old commits. I'm guessing it's because they just so happen to be matched by the parsing. This is fine, I reckon.

But the changelog does not contain all of the recent commits already made using commitizen, only `fix` appear in the changelog. 

Using the DEBUG env var shows things like:

```
semantic_release:history:logs 'ignored commit' UnknownCommitMessageStyleError('Unable to parse the given commit message: revert(automation): remove outdated playbook\n') +1.1ms
```

```
semantic_release:history:logs 'ignored commit' UnknownCommitMessageStyleError('Unable to parse the given commit message: improvement(apiserver): pATCH on accounts now allow for updating protecte\n\nWhen instagram accounts are authenticated as Business accounts, their IDs change. This change allows\nsm_account_id changes for instagram accounts.\n\nre #5230\n') +1.9ms
```

Others are successful on the parsing:

```
semantic_release:history:parser_angular "parse_commit_message -> (0, refactor, apiserver,connector,sorter, ('simplify logic on detecting business instagram accounts', 'As Basic Display API is not supported, the need for detecting what kind of API the account should use has been simplified. ', ''))" +259.0us
```

_But they still don't show in the changelog_.

For test, I used it for the first time - bumping version, etc. Then made a few test commits and did the same. Not all commits show.

My setup.cfg is barebones:

```
[semantic_release]
version_variable = src/sm2/__init__.py:__version__
branch = develop  # testing
```

nicholasamorim:
Customization of changelog can be done via changelog_sections var in configuration.

Fixes #193

### DIFF
--- a/semantic_release/history/logs.py
+++ b/semantic_release/history/logs.py
@@ -17,16 +17,6 @@ LEVELS = {
     3: 'major',
 }
 
-# Sections which will be shown in the Markdown changelog.
-# This is NOT related to supported commit types.
-CHANGELOG_SECTIONS = [
-    'feature',
-    'fix',
-    'breaking',
-    'documentation',
-    'performance',
-]
-
 
 def evaluate_version_bump(current_version: str, force: str = None) -> Optional[str]:
     """
@@ -173,8 +163,13 @@ def markdown_changelog(version: str, changelog: dict, header: bool = False) -> s
         # Add a heading with the version number
         output += '## v{0}\n'.format(version)
 
-    for section in CHANGELOG_SECTIONS:
-        if not changelog[section]:
+    # Sections which will be shown in the Markdown changelog.
+    # This is NOT related to supported commit types.
+    changelog_sections = config.get("semantic_release", "changelog_sections")
+    changelog_sections = [s.strip() for s in changelog_sections.split(',')]
+
+    for section in changelog_sections:
+        if section not in changelog or not changelog[section]:
             # This section does not have any commits
             continue
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -97,6 +97,25 @@ Default: ``semantic-release <semantic-release>``
   If you are using the built-in GitHub Action, this is always set to
   ``github-actions <actions@github.com>``.
 
+Changelog
+=========
+
+``changelog_sections``
+-----------------------
+Comma-separated list of sections to display in the changelog.
+They will be displayed in the order they are given.
+
+The available options are:
+
+* ``feature``
+* ``fix``
+* ``documentation``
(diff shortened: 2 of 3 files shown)


semantic_release/history/logs.py
```diff
@@ -19,7 +19,7 @@
 
 # Sections which will be shown in the Markdown changelog.
 # This is NOT related to supported commit types.
-CHANGELOG_SECTIONS = [
+DEFAULT_CHANGELOG_SECTIONS = [
```
AlphaMycelium, line 22: Why not use `defaults.cfg`?

semantic_release/history/logs.py
```diff
@@ -173,7 +173,14 @@ def markdown_changelog(version: str, changelog: dict, header: bool = False) -> s
...
-    for section in CHANGELOG_SECTIONS:
+    custom_changelog_sections = config.get("semantic_release", "changelog_sections", fallback="")
+    if custom_changelog_sections:
+        debug('markdown_changelog(custom_changelog={})'.format(custom_changelog_sections))
+        changelog_sections = [s.strip() for s in custom_changelog_sections.split(',')]
+    else:
+        changelog_sections = DEFAULT_CHANGELOG_SECTIONS
+
+    for section in changelog_sections:
         if not changelog[section]:
```
AlphaMycelium, line 184: Maybe add a check to skip invalid sections:

```suggestion
        if section not in changelog or not changelog[section]:
```

semantic_release/history/logs.py
nicholasamorim: Ah, I'm just not familiar with the code, will change.

semantic_release/history/logs.py
nicholasamorim: Will do

AlphaMycelium:
Ah, the release workflow failed because I merged the PR, so it doesn't have permission to update `master` because it is protected. I'll leave merging to @relekang next time.

relekang:
Hmm, I guess the same would happen for me. There is no difference on our rights to push to master. We might need to add github actions as an excemption 🤔

relekang:
I removed the protection on master for now.

PR #208, opened 2020-04-07
Automate creation of API docs

AlphaMycelium:
Make the API docs section of the documentation automatically-generated using `sphinx-apidoc`, meaning that new modules and packages are added automatically. This is added as an event handler in conf.py so that it always runs before building, and [should work on ReadTheDocs](https://github.com/readthedocs/readthedocs.org/issues/1139#issuecomment-187884185).


relekang:
Nice 🙌

PR #214, opened 2020-04-13
Beautify code automatically instead of checking with flake8

AlphaMycelium:
Instead of running `flake8` and `isort` as checks, use `black` and `isort` to format code automatically, via GitHub Actions. Any changes are committed as `github-actions` and pushed.

## Advantages

- Contributors don't need to think about code style because any problems will be fixed automatically.
- Styling is the best it can be, not just anything which follows pep8.

## Disadvantages

- A lot of automated commits will make the history a bit cluttered.
  - Looking at repositories such as [unibeautify](https://github.com/Unibeautify/unibeautify/commits/master), this seems to be an acceptable thing on GitHub, but it's up to @relekang to decide.
  - To reduce the number of commits made we could run this on a schedule (e.g. every week or month).
  - Additionally, contributors can be asked to run `isort -y -rc semantic_release tests && black .` before committing.

---

@relekang, I understand that this may be something you don't want, if so, feel free to close this.


relekang:
I am all for automatic fixing. Let's make 🤖 do the work. However, I am wondering how this will affect the automatic release. Maybe we should make this check run in the regular action when pushed to master and abort release if they fail?

AlphaMycelium:
Hmmm. I'm trying to avoid all these "skipped" sections at the bottom of pull requests. What I'll do is duplicate the main workflow to have one version for PRs, and another which runs on master pushes.

AlphaMycelium:
@relekang I've just realised something: Since the workflow is running for a certain commit, the release step will still checkout the commit from before it was beautified. This shouldn't cause a problem since the meaning of the code isn't changed, but what it means is that we may as well run the beautification in parallel from a separate workflow - which also allows us to skip it completely if no Python files have been modified.

PR #215, opened 2020-04-13
Store requirements in setup.py

AlphaMycelium:
Remove the requirements directory and instead store all required  libraries directly inside setup.py. 

Development, testing and docs dependencies are added as optional extras, and can be installed with:
```
pip install .[dev,test,docs]
```

| Extras | Contents |
| --- | --- |
| `dev` | Tools such as `tox` and `isort` |
| `test` | `pytest` and related libraries |
| `docs` | Sphinx |


AlphaMycelium:
@relekang This seems to have broken something with the Read the Docs build, since I can't access the build logs you'll have to take a look.

relekang:
```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements/base.txt'
```

AlphaMycelium:
Ah, I fixed it by creating `.readthedocs.yml` with the correct settings.

PR #220, opened 2020-04-26
Support easier use of PyPI API tokens

### LINKED ISSUE #213: Support authenticating to PyPI with an API token
## Description

Support PyPI authentication using an API token.

## Use cases

- Password is not stored on CI servers
- Scope can be limited to only access the project being uploaded

## Implementation

Set `__token__` as the username and send the token as the password, as described [here](https://pypi.org/help/#apitoken).

This is already possible using the current system, but it would be nicer to be able to just set the token in an environment variable such as `PYPI_TOKEN`, and have the username filled in automatically.

AlphaMycelium:
- Allow setting the environment variable `PYPI_TOKEN`.
- Tokens are preferred over username and password if both are set.

Fixes #213

### DIFF
--- a/action.sh
+++ b/action.sh
@@ -2,6 +2,7 @@
 
 # Copy inputs into correctly-named environment variables
 export GH_TOKEN="${INPUT_GITHUB_TOKEN}"
+export PYPI_TOKEN="${INPUT_PYPI_TOKEN}"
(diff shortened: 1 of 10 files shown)


BjoernLudwigPTB:
Yes, awesome, thanks for that! I found it a little strange, when I introduced such a generic environment variable like `__token__` into our pipeline on CircleCI and will dump it right away. :smile:

PR #226, opened 2020-04-30
Close stale questions automatically

AlphaMycelium:
Any issues which are labelled as a question will be marked as `stale` after 4 days with no activity, and then closed after a further 3 days. This doesn't affect other types of issues.


relekang:
I think the time should be at least a couple of weeks, so that things that happen when we have vacation don't get auto closed.

AlphaMycelium:
I have opened [this issue](https://github.com/actions/stale/issues/56) in relation to only closing issues which have been responded to. In the meantime, I've increased the time to two weeks.

PR #232, opened 2020-05-08
Add link to view commit range in changelog

### LINKED ISSUE #218: Include the commit range since the last release into changelog
## Description

We could add a section at the bottom of the changelog that links to the commit range on GitHub for a certain release since the previous release with heading like "Latest changes".

I got the idea from [another project](https://github.com/klaussinani/ao/releases/tag/v6.9.0) and find it very useful.

Maybe we would want to introduce a new cli parameter `--include-commit-range` for that to be an optional behaviour.

## Use cases

Including the commit range would allow developers to quickly get hands on the associated commits for a certain release.

## Possible implementation

The GitHub link we get almost as a side effect since it is of the form.

```
https://github.com/<user_or_oranization>/<project>/compare/<previous-version-tag>...<current-version-tag>
```

This we could easily include in the generated changelog.

AlphaMycelium:
Display a link in the changelog to the page comparing commits between this release and the previous one.

This is configured using the config option `compare_link`, and is off by default since GitHub displays a "compare" menu next to the release for the same purpose:

image

Fixes #218

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -136,6 +136,17 @@ The available options are:
 
 Default: `feature, fix, breaking, documentation, performance`
 
+.. _config-compare_link:
+
+``compare_link``
+----------------
+When set to true, add a link to view the commit range since the previous
+release, on GitHub. Only appears when using :ref:`cmd-publish`.
+
+.. note::
+  If you are using a different HVCS, the link will not be included
+  regardless of this option.
+
 Distributions
 =============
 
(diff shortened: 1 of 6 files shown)


AlphaMycelium:
@relekang Hmm, I don't think [this Actions error](https://github.com/relekang/python-semantic-release/actions/runs/99160368) has anything to do with the actual code. I guess it's easiest to just wait for another commit to be pushed to run it again.

PR #234, opened 2020-05-11
Capture logging output in tests

AlphaMycelium:
Adapt the CLI tests so that pytest can capture the log output, and display it if there is a failure. This helps with debugging.


relekang:
I think we loose some value by not running the click parser in the tests. 🤔

AlphaMycelium:
@relekang I kept the click parser in tests such as `test_force_major`, which are only checking that CLI arguments translate into the correct function call. We can always create more of those to test the parsing separately from other logic.

Issue #175, opened 2020-02-18
Bump versions in multiple files?


ttimbers:
It would be nice to bump versions in multiple files. For example, in a package that uses `pyproject.toml` (and thus has the version stored there) and that also has the version stored in `package/__init__.py`. Maybe this could work by setting multiple filenames in `version_variable`?

relekang:
> Maybe this could work by setting multiple filenames in version_variable?

Good idea, maybe just comma separate it?

ttimbers:
Are you suggesting I should just try it and see if it works, or is that an implementation suggestion?

relekang:
It is an implementation suggestion.

alandtse:
As a note, I believe with Python 3.x[ libraries can source the version info from pyproject.toml](https://github.com/python-poetry/poetry/issues/273#issuecomment-570999678) so you actually don't need this feature.

PR #246, opened 2020-06-14
Bump versions in multiple files

kalekundert:
This PR adds the ability to bump versions in multiple files as described in #175.  It also adds the ability to identify version numbers using arbitrary regular expressions (in addition to variable names).  Below is an example of a `pyproject.toml` file that uses the new features:

```toml
[tool.semantic_release]
# Both settings can be either lists or strings, so this is a backwards-compatible change.
version_variable = [
    'path/to/file1:var1',
    'path/to/file2:var2'
]
version_pattern = [
    'path/to/file3:reg(ex)',  # The group identifies the version number to bump.
    'path/to/file4:{version}'  # {version} is replaced with the standard version regex, e.g. matching 1.2.3
]
```
One caveat worth discussing is that this implementation only allows multiple files to be specified in `pyproject.toml`, not in `setup.cfg`.  This is due to the fact that the `setup.cfg` file format does not support composite data types such as lists; every setting has to be a string.  In many cases you can use a delimiter (e.g. `,` as suggested in #175) to split a setting into a list, but this is not an option for `version_pattern`, which may reasonably contain any delimiter.  I gave this a lot of thought, and ultimately I think that requiring TOML is the most correct and future-proof way to specify multiple path/pattern pairs.

This PR also refactors the global `config` object.  Previously, this object was a `ConfigParser` (i.e. the class used to parse `setup.cfg`), which meant that values parsed from TOML had to be converted to strings before they could be added to the data structure.  This architecture made it impossible to use lists from TOML.  Now the global `config` object is a regular dictionary with values taken from `setup.cfg` and/or `pyproject.toml`.

Task list:

- [X] Implementation
- [X] Tests
- [x] Documentation


kalekundert:
Ok, I think it's ready to go now.  Let me know if it needs any changes!

semantic_release/history/__init__.py
```diff
@@ -114,49 +190,40 @@ def get_previous_version(version: str) -> Optional[str]:
...
+
+def load_version_patterns() -> List[VersionPattern]:
+    """
+    Create the `VersionPattern` objects specified by the config file.
+    """
+    patterns = []
+
+    def iter_fields(x):
+        if not x: return
+        yield from x if isinstance(x, list) else [x]
```
relekang, line 216: Can we please change this so it supports comma separated lists as well?
```suggestion
        yield from x if isinstance(x, list) else x.split(",")
```

kalekundert: Done.  I realized that I could use `csv.reader()` to make it possible to escape commas, so I did that.

PR #252, opened 2020-07-24
Correct docs for changelog command to reflect correct output stream

### LINKED ISSUE #250: Apparently misleading documentation for changelog regarding output stream
The current documentation states:

> changelog¶
>When executed this command will print the changelog to stdout.

Apparently the commands prints to stderr, at least when used with the parameter `--unreleased`. If this is the case, we should adapt/correct [the docs](https://python-semantic-release.readthedocs.io/en/latest/commands.html?highlight=stdout#changelog).

BjoernLudwigPTB:
This resolves #250. In case it is the case that in general the following command prints to stderr, this now is reflected in the docs:

```shell
semantic-release --unreleased changelog
```

In case this is conditional, I am very happy to properly document the feature. Just give me a sign and I will dive deeper into the actual behaviour.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -7,7 +7,7 @@ Commands
 
 changelog
 ^^^^^^^^^
-When executed this command will print the changelog to stdout.
+When executed this command will print the changelog to stderr.
 
 If the option ``--post`` is used then the program will check if
 there is a authentication token configured for your vcs provider


AlphaMycelium:
This appears to have changed in 15b1f650f29761e1ab2a91b767cbff79b2057a4c, since [`click-log` uses stderr by default](https://github.com/click-contrib/click-log/blob/0d72a212ae7a45ab890d6e88a690679f8b946937/click_log/core.py#L46-L55). Perhaps it would be better to output the changelog on stdout (using a standard `print`), separate from log messages.

BjoernLudwigPTB:
> This appears to have changed in [15b1f65](https://github.com/relekang/python-semantic-release/commit/15b1f650f29761e1ab2a91b767cbff79b2057a4c), since [`click-log` uses stderr by default](https://github.com/click-contrib/click-log/blob/0d72a212ae7a45ab890d6e88a690679f8b946937/click_log/core.py#L46-L55). Perhaps it would be better to output the changelog on stdout (using a standard `print`), separate from log messages.

If it would be a help, I could correct and extend my PR.

AlphaMycelium:
I changed it in #254, I'll leave it up to @relekang to decide which PR gets merged.

relekang:
Closing this since #254 got merged.

PR #254, opened 2020-07-24
Send changelog to stdout

### LINKED ISSUE #250: Apparently misleading documentation for changelog regarding output stream

AlphaMycelium:
This appears to have changed in 15b1f650f29761e1ab2a91b767cbff79b2057a4c, since [`click-log` uses stderr by default](https://github.com/click-contrib/click-log/blob/0d72a212ae7a45ab890d6e88a690679f8b946937/click_log/core.py#L46-L55).

Fixes #250, this PR is an alternative to #252.

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -161,7 +161,8 @@ def changelog(*, unreleased=False, noop=False, post=False, **kwargs):
         log = generate_changelog(current_version, None)
     else:
         log = generate_changelog(previous_version, current_version)
-    logger.info(markdown_changelog(current_version, log, header=False))
+    # print is used to keep the changelog on stdout, separate from log messages
+    print(markdown_changelog(current_version, log, header=False))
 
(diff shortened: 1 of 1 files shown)


BjoernLudwigPTB:
Just switching from stderr to stdout might break some CI pipelines. If they relied on the changelog being printed on stderr, they might miss their expected output. An option would be to copy the output to both streams.

AlphaMycelium:
It was originally sent to stdout until the commit I mentioned, where I didn't notice the change to stderr, so I guess it is a bug. Perhaps we should create a test for this.

PR #256, opened 2020-07-29
Create getting started instructions

### LINKED ISSUE #255: Need Getting Started documentation
The current documentation is written in a way that probably seems logical to users familiar with the project.  However, for new users, a Getting Started section would be most helpful.

For example, I have my `setup.cfg` configured, my `__init__.py` has the semver properly.  If I run `semantic-release version --patch`, the version is updated correctly.

But, how do I trigger a version update in my manual build process? 

I tried this log message:
```
commit 40d84c691162937771793b3071f581c9fb6a1320 (HEAD -> feature/monorepo)
Author: Mike Crowe <drmikecrowe@gmail.com>
Date:   Mon Jul 27 07:07:02 2020 -0400

    :bug: Consolidating main
```
Shouldn't running `semantic-version version` work now?

Is the git log correct for the emoji parser?
Is the fact I'm on a branch an issue?
What needs to happen in git?  Do I need to be on a branch?  Need to tag anything?

This is a quasi-monorepo, where this python project is in a `device/` subdirectory (and other packages are in `/packages`).  Is the fact it's in a subdirectory an issue?

AlphaMycelium:
Fixes #255

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
 Python Semantic Release
-=======================
+***********************
 
 |Test Status| |PyPI Version| |conda-forge version| |Read the Docs Status|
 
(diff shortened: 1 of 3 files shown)


relekang:
Why did we remove all the info from the readme?

AlphaMycelium:
@relekang `:ref:`, `..seealso` etc don't work in the README as it's not built using Sphinx. Adding a prominent link pointing people towards the docs could help? Or we could rewrite the getting started section using regular links so it could be displayed on GitHub.

relekang:
A link would be nice, I think we should also keep a shorter version like

# Setup

>```
>pip install python-semantic-release
>semantic-release --help
>```
>Read more about the setup in our [getting started guide](...)

and maybe also a note that it exists as a github action in the readme?

What do you think?

AlphaMycelium:
Sounds good, I'll add that soon :+1:

PR #259, opened 2020-09-09
Support multiline version_pattern matching by default

### LINKED ISSUE #258: Unable to specify version_pattern with start/end of line (^ or $)
## The problem

`version_pattern` does not allow for the use of `^` or `$` to signify the start or end of a line. This means that some unwanted matches might be found (examples below)

## Expected behavior

Using `^` or `$` would match the start or end of a line in the target file, as in `multiline` regular expressions.

## Additional context

`pyproject.toml`
```toml
[tool.poetry]
version = "0.1.0"

[tool.semantic_release]
version_pattern = 'pyproject.toml:version = "{version}"'

[tool.pytest.ini_options]
minversion = "6.0"
```

Running `semantic-release --noop version` results in the following:
```console
$ semantic-release --noop version
Creating new version
error: found conflicting versions: '0.1.0', '6.0'
```

It appears this is due to the use of `re.finditer` in [VersionPattern.parse](https://github.com/relekang/python-semantic-release/blob/34a2b2e91c2439fec7f08bfc628d7f470ef2fdfe/semantic_release/history/__init__.py#L74).

## Workarounds

There is a simple way to work around this limitation: specify the inline multiline flag `(?m)` somewhere in the pattern string, e.g.:

```toml
version_pattern = 'pyproject.toml:version = "{version}"(?m)'
```

However, given the use cases for this tool and the places where the version is likely to be, it seems like it might be good to turn on the multiline flag by default.

jmaroeder:
Closes #258 

This enables `re.MULTILINE` matching for `version_pattern` specifiers by default.

It could be argued that this is actually not a fix but a breaking change, as technically using `^` or `$` would match the start/end of the entire file contents with the current version of code, but in practice I doubt this is likely to be the case.

### DIFF
--- a/semantic_release/history/__init__.py
+++ b/semantic_release/history/__init__.py
@@ -24,9 +24,9 @@ class VersionPattern:
     """
     Represent a version number in a particular file.
 
-    The version number is identified by a regular expression.  Methods are 
-    provided both the read the version number from the file, and to update the 
-    file with a new version number.  Use the `load_version_patterns()` factory 
+    The version number is identified by a regular expression.  Methods are
+    provided both the read the version number from the file, and to update the
+    file with a new version number.  Use the `load_version_patterns()` factory
     function to create the version patterns specified in the config files.
     """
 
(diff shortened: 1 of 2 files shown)


relekang:
Thanks 👍

PR #263, opened 2020-09-16
Check commit logs with commitlint in CI

browniebroke:
The contributing guide says that the project should itself follow the Angular commit convention, but there is nothing to enforce it AFAIK.

I had a similar problem on a project where I'm using `python-semantic-release` and I've added a Github action to test it on CI, you might find it useful too.

I also recently added [a pre-commit hook](https://github.com/alessandrojcm/commitlint-pre-commit-hook) to get even earlier feedback, let me know if you're interested.


danth (approved):
Currently we use squash commits to allow the person merging to change the commit message, however this seems like a much more elegant solution.

PR #266, opened 2020-09-17
feat: Generate `changelog.md` file

### LINKED ISSUE #27: Generation of changelog.md

browniebroke:
Fixes #27

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -32,6 +32,7 @@ This will not push anything to any remote. All changes are local.
 
 Publish will do a sequence of things:
 
+#. Update changelog file.
 #. Run :ref:`cmd-version`.
 #. Push changes to git.
 #. Run :ref:`config-build_command` and upload the created files to PyPI.
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -278,6 +278,24 @@ take any of the following values as keyword arguments:
 
 You can should use ``**kwargs`` to capture any arguments you don't need.
 
+``changelog_file``
+------------------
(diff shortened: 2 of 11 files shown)


browniebroke:
The documentation is missing at this point. It would be nice to check that the approach makes sense before I do so.

Overall, this how it's supposed to work, it will:

- Check if a new version should be released, if not stop here, otherwise continue.
- Update `CHANGELOG.md` and replace the placeholder `<!--next-version-placeholder-->` with changelog from the version being released
- Do `git add` for the `CHANGELOG.md` file
- Proceed as currently with updating version in source control, git add, commit, push, etc...

What are the expected behaviour when the `CHANGELOG.md` file doesn't exist or when it doesn't have the placeholder? Should we just skip this step and log a warning?

browniebroke:
Actually, I've update the docs. 

I realised while writing it that if the changelog file doesn't exist, we can have a better default and create it, so I've changed the code accordingly. However if the placeholder isn't in the file, probably safer to skip this step.

I'm open to suggestions to alternative solution that work without placeholders (or give it a better name).

browniebroke:
Nearly there: https://github.com/browniebroke/django-codemod/blob/master/CHANGELOG.md

The commit hash aren't linked automatically like in the GH release... I guess we could generate the links but would need to be working with gitlab as well.

browniebroke:
You can see the results of the following commits in releases v0.17.3, v0.17.4 and v0.17.5 here: https://github.com/browniebroke/django-codemod/blob/master/CHANGELOG.md

relekang (approved):
Thanks 🙌

PR #281, opened 2020-12-03
feat(logs): include scope in changelogs

wyardley:
When the scope is set, include it in changelogs, e.g.,
`"feat(x): some description"` becomes:
`* **x**: some description`

in the changelog. This is similar to how the Node semantic release (and conventional-changelog-generator) generate changelogs. If scope is not given, it's omitted.

Add a new config parameter changelog_scope to disable this behavior when set to `False`

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -297,6 +297,13 @@ will be updated.
 
 Default: ``<!--next-version-placeholder-->``.
 
+``changelog_scope``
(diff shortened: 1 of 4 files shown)


wyardley:
Totally open to adjusting test cases, switching the default, or other feedback.

At some point, I'd also like to add a separate change to make PR links clickable in the changelog when present.

Issue #123, opened 2019-05-29
Enhancement: Support Github URLs


rykard95:
As a developer, I would like to tag repos in my personal github, and noth github.com.

relekang:
PRs are welcome 🙂

relekang:
I am not sure what you mean by personal github. Please update the issue and re-open if you still want this.

Issue #277, opened 2020-11-27
Support GitHub Enterprise


prabu28:
push_new_version(, auth_token='XXXXXXX-XXXXXX-XXXXXXXXX', owner='appowner', name='MyAPI', branch='dev', domain=**'github.com'**)

is there a way to make publish work for github enterprise  ? some thing like **github.mycompany.com**
 i tried setting domain = github.mycompany.com in setup.cfg but that dint help .

danth:
Currently, Python Semantic Release does not support this.

If the GitHub Enterprise API uses the same endpoints as `api.github.com`, only under a different domain name, we can add a configuration option to set `API_URL` in `hvcs.py`.

prabu28:
> Currently, Python Semantic Release does not support this.
> 
> If the GitHub Enterprise API uses the same endpoints as `api.github.com`, only under a different domain name, we can add a configuration option to set `API_URL` in `hvcs.py`.

thanks for the info  @danth

wyardley:
I was looking at implementing this, using `hvcs_domain` as an option that would work for both GHE and Gitlab, but it seems that Gitlab uses `CI_SERVER_HOST` for this already.

https://github.com/relekang/python-semantic-release/blob/af4cd6a283ac8184637c438fc11f35aed7640cee/semantic_release/hvcs.py#L270

Also, presumably it may be necessary to use the domain from hvcs.py for the links in the changelog?

I'm still thinking it's cleanest to allow `hvcs_domain` to be configurable. I have some work in progress on this, and can push it and work on tests after we see what happens with #282.

May be a more tidy / efficient way to do some of this, but this is what I have thus far.
https://github.com/relekang/python-semantic-release/compare/master...wyardley:wyardley-hvcs-domain?expand=1

wyardley:
@prabu28: do you know if the API hostname is consistent? e.g., if you have a vanity domain of github.example.com for GHE, will you use github.example.com, or api.github.example.com for the API? It seems like some instances have the API hostname, but that maybe it should work without it, and that might be the safest / most consistent?

prabu28:
> @prabu28: do you know if the API hostname is consistent? e.g., if you have a vanity domain of github.example.com for GHE, will you use github.example.com, or api.github.example.com for the API? It seems like some instances have the API hostname, but that maybe it should work without it, and that might be the safest / most consistent?

hi @wyardley  https://github.mycomany.com/api/v3 is our api url

PR #282, opened 2020-12-05
feat(changelog): add PR links in markdown

wyardley:
GitHub release notes automagically link to the PR, but changelog markdown doesn't. Replace (#123) at the end of a message with a markdown link.

### DIFF
--- a/semantic_release/changelog/changelog.py
+++ b/semantic_release/changelog/changelog.py
@@ -1,6 +1,30 @@
+import re
+
 from typing import Iterable, Optional
 
 from ..settings import config
+
+
+def add_pr_link(owner: str, repo_name: str, message: str) -> str:
+    """
+    GitHub release notes automagically link to the PR, but changelog markdown
+    doesn't. Replace (#123) at the end of a message with a markdown link.
+    """
+
+    pr_pattern = re.compile(r"\s+\(#(\d{1,8})\)$")
+    match = re.search(pr_pattern, message)
+
+    if match:
+        pr_number = match.group(1)
+        url = (
+            f"https://gitlab.com/{owner}/{repo_name}/-/merge_requests/{pr_number}"
+            if config.get("hvcs") == "gitlab"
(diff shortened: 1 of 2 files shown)


wyardley:
Open to suggestions on style / formatting here. I decided to do the f string for the table formatting in two lines, and to put the spaces at the beginning of the next line for the tests, but not super married to either way of doing it.

Similarly, could easily make the regex part little terser or change stuff there if it would be preferred.

semantic_release/changelog/changelog.py
```diff
@@ -1,8 +1,32 @@
...
+    """
+
+    pr_pattern = re.compile(r'\s+\(#(\d{1,8})\)$')
+    match = re.search(pr_pattern, message)
+
+    if match:
+        pr_number = match.group(1)
+        url = (
+            f"https://gitlab.com/{owner}/{repo_name}/-/merge_requests/{pr_number}"
+            if config.get("hvcs") == "gitlab"
```
danth, line 21: Could you add a test for GitLab links?

wyardley: Sure!
Is a unit test for `add_github_link()` good (added), or do you also want me to add a test for `test_markdown_changelog()` / `test_changelog_table()` that sets hvcs to gitlab? (note: there is still no test coverage for the one that does the same thing for commit hashes).

I don't use GitLab much, but did verify that this _should_ be the right link pattern, e.g.,
https://gitlab.com/muttmua/mutt/-/merge_requests/120

wyardley: Update: added a separate test for gitlab `test_markdown_changelog()` as well as the unit test; that will also test `get_hash_link()` indirectly.

wyardley:
Also, we may want to (fairly soon) update the gitlab ones to follow the same hostname as for the API, assuming that's the correct behavior, but that would probably go along with the changes I mentioned in #277

danth (approved):

PR #283, opened 2020-12-05
style(settings): alphabetize boolean settings

wyardley:
trivial: A few settings were not in alphabetical order.


wyardley:
From the context, I was guessing that these were intended to be alphabetical, but feel free to close if they're in some other order.

PR #284, opened 2020-12-05
refactor(history): move changelog_scope default

wyardley:
Some cleanup from #281

- Move the default for changelog_scope from inline to defaults.cfg
- Add missing `.. _config-changelog_scope:` header in docs


wyardley:
I hadn't realized where these defaults are set, but this probably lines up better with the rest of the default settings.

PR #285, opened 2020-12-05
WIP: feat(cli): add build_dist setting

wyardley:
Add a new setting`build_dist` (defaulting to true). If set, build_command will not be run.

While it's always possible to set build_command to "true", this is useful for the case where a dist is not being built in the first place.


wyardley:
This just needs tests, but wanted to make sure this was going in the right direction first. Any thoughts about the best (and easiest) way to test? Would be easier to unit test if the conditional were in https://github.com/relekang/python-semantic-release/blob/master/semantic_release/dist.py, but based on the existing code, looks like this should live in CLI?

I thought about just skipping the build command if it's set to an empty string, or the string "False", but this seems a bit cleaner and in line with the way the rest of the options are setup.

danth:
What happens if `build_dist` is false, but we expect a file because `upload_to_pypi` or `upload_to_release` is enabled?

wyardley:
> What happens if build_dist is false, but we expect a file because upload_to_pypi or upload_to_release is enabled?

Yeah, I was wondering if this parameter, if added, should also disable / bypass (or at least contradict / raise an error with) some other settings

danth:
It is best if this setting disables all uploads, giving a warning if an upload explicitly set to true will not happen.

wyardley:
Interesting - based on what I can see, with the settings:
```
upload_to_release = false
upload_to_pypi = false
```
`build_dist()` should _already_ not be getting called, and this would thus be unnecessary as an additional feature. IIRC, this wasn't working as I had expected before, but will give it another shot to confirm.

wyardley:
Confirmed. not needed, _but_ I will put in a separate, more limited, PR for a separate corner case that's related.

PR #289, opened 2020-12-12
feat(hvcs): add hvcs_domain config option

wyardley:
While Gitlab already has an env var that should provide the vanity URL, this supports a generic 'hvcs_domain' parameter that makes the hostname configurable for both GHE and Gitlab.

This will also use the configured hostname (from either source) in the changelog links

Fixes: #277

### DIFF
--- a/semantic_release/hvcs.py
+++ b/semantic_release/hvcs.py
@@ -26,6 +26,10 @@ class Base(object):
         raise NotImplementedError
 
     @staticmethod
+    def api_url() -> str:
+        raise NotImplementedError
+
+    @staticmethod
     def token() -> Optional[str]:
         raise NotImplementedError
 
@@ -86,7 +90,7 @@ class TokenAuth(AuthBase):
 class Github(Base):
     """Github helper class"""
 
-    API_URL = "https://api.github.com"
+    DEFAULT_DOMAIN = "github.com"
     _fix_mime_types()
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


wyardley:
Functionally, this seems to work, however, unsure about a couple things:
1) Having some trouble testing this properly (see failing tests); once I sort that out, I can add more tests for the Github case, as well as add tests for the changelog generation.
2) Not sure if this should still be in all caps, since it's not a constant? And should this be in an `__init__()` or similar function for the `Github` / `Gitlab` classes and then change it to `self.BASE_URL` or similar?
3) I'm assuming it's better to call `Github.domain()` vs directly referencing `Github.BASE_URL` from outside the class, right?

We would also want to verify that the vanity URL is always used for the PR links and so on, though I think it should be. Let me know if there are other places that URLs should follow the configured ones as well.

wyardley:
I'm thinking that it's likely not as performant to call these methods that call `config.get()` so frequently or use these static methods? would appreciate some guidance on how best to structure this it if you've got any ideas. @relekang @danth

semantic_release/hvcs.py
```diff
@@ -176,8 +196,8 @@ def get_release(cls, owner: str, repo: str, tag: str) -> int:
         :return: ID of found release
         """
         response = requests.get(
-            f"{Github.API_URL}/repos/{owner}/{repo}/releases/tags/{tag}",
-            auth=Github.auth(),
+            f"{Github.api_url()}/repos/{owner}/{repo}/releases/tags/{tag}",
+            headers={"Authorization": "token {}".format(Github.token())},
```
relekang, line 200: Please change this back to use `auth=Github.auth()`

relekang (changes requested):
Nice, seems just one little thing got mixed up in fixing of conflicts.

semantic_release/hvcs.py
wyardley: derp, sorry, I knew something was a little funny here.

Fixed.
However, the test from #296 (`GithubReleaseTests.test_should_post_changelog_using_netrc`) seems to fail for me locally both on trunk and on this branch; not sure if it will pass here.

wyardley: And let me know if you have a preferred / better way to implement this; happy to rework it some if needed.

PR #290, opened 2020-12-12
fix(cli): skip remove_dist where not needed

wyardley:
Skip removing dist files when `upload_pypi` or `upload_release` are not set.

This way, people don't have to explicitly set `remove_dist()` to `False` when `build_dist()` has not run.

I _think_ this should be the correct behavior, and this seemed like a relatively simple way to accomplish it.

Open to suggestions on a) other / better ways to accomplish this and b) best / simplest ways to add a unit test for this.

Alternately, I could move the `if upload_pypi` and `if upload_release` conditionals inside that outer block:
https://github.com/relekang/python-semantic-release/blob/1b386c6d78772fd09351697c83fd4e39dfae1821/semantic_release/cli.py#L265
and then publish the changelog, however, I'm not sure whether that would be the wrong order of operations

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -298,7 +298,7 @@ def publish(**kwargs):
             logger.info("Uploading to HVCS release")
             upload_to_release(owner, name, new_version, dist_path)
         # Remove distribution files as they are no longer needed
-        if remove_dist:
+        if (upload_pypi or upload_release) and remove_dist:
             remove_dists(dist_path)
 
         logger.info("New release published")


wyardley:
```diff
diff --git a/semantic_release/cli.py b/semantic_release/cli.py
index c570138..8b643a2 100644
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -270,13 +270,21 @@ def publish(**kwargs):
                 remove_dists(dist_path)
             build_dists()
 
-        if upload_pypi:
-            logger.info("Uploading to PyPI")
-            upload_to_pypi(
-                path=dist_path,
-                # If we are retrying, we don't want errors for files that are already on PyPI.
-                skip_existing=retry,
-            )
+            if upload_pypi:
+                logger.info("Uploading to PyPI")
+                upload_to_pypi(
+                    path=dist_path,
+                    # If we are retrying, we don't want errors for files that are already on PyPI.
+                    skip_existing=retry,
+                )
+
+            # Upload to GitHub Releases
+            if upload_release and check_token():
+                logger.info("Uploading to HVCS release")
+                upload_to_release(owner, name, new_version, dist_path)
+            # Remove distribution files as they are no longer needed
+            if remove_dist:
+                remove_dists(dist_path)
 
         if check_token():
             # Update changelog on HVCS
@@ -288,14 +296,6 @@ def publish(**kwargs):
         else:
             logger.warning("Missing token: cannot post changelog to HVCS")
 
-        # Upload to GitHub Releases
-        if upload_release and check_token():
-            logger.info("Uploading to HVCS release")
-            upload_to_release(owner, name, new_version, dist_path)
-        # Remove distribution files as they are no longer needed
-        if remove_dist:
-            remove_dists(dist_path)
-
         logger.info("New release published")
 
     # else: Since version shows a message on failure, we do not need to print another.
```

alternative method

danth:
The changelog needs to be posted before `upload_to_release` - uploads require a release to attach to. Hence, your alternative method will not work.

semantic_release/cli.py
```diff
@@ -292,9 +292,10 @@ def publish(**kwargs):
         if upload_release and check_token():
             logger.info("Uploading to HVCS release")
             upload_to_release(owner, name, new_version, dist_path)
-        # Remove distribution files as they are no longer needed
-        if remove_dist:
-            remove_dists(dist_path)
+
+            # Remove distribution files as they are no longer needed
+            if remove_dist:
```
relekang, line 297: Instead of moving this in here change it to `if (upload_pypi or upload_release)  and remove_dist:`

relekang (changes requested):

semantic_release/cli.py
wyardley: 👍 
updated

PR #291, opened 2020-12-13
fix(version_variable): fixing version extraction/setting from/in toml files via version_variable

### LINKED ISSUE #275: Version bump tries to bump all matching of `version =` in pyproject.toml
## The problem

With version 7.3.0:

In pyproject.toml, the version of the main package is defined as `version = 1.2.3` but some dependencies can be defined as such `pyarrow = {version = "0.17.0", extras = [ "pandas" ] }` and then semantic release is either bumping both `version = ` the the computed version or trowing an error such as:

```
semantic-release version --patch
Creating new version
error: found conflicting versions: '1.2.3', '0.17.0'
```

## Expected behavior

semantic release only bump the root version of the package and nothing in the dependencies section.

## Additional context

Here is an example of pyproject.toml that you can use to reproduce
```
[tool.poetry]
name = "potatoes"
version = "1.2.3"
description = "I love potatoes"
authors = ["Potatoes <potatoes@potatoes.com>"]

[tool.poetry.dependencies]
python = "^3.8"
pyarrow = {version = "0.17.0", extras = [ "pandas" ] }

[tool.semantic_release]
version_variable = "pyproject.toml:version"
upload_to_pypi = false
commit_message = "Version generated by python-semantic-release [ci skip]"
```

Cielquan:
I made a quick fix for #275 to correctly handle toml files like `pyproject.toml` holding version strings.

I personally would rewrite the whole `history/__init__.py` module as it is intended to work with regex only and my solution therefore is a bit clunky implemented.

I would exchange `toml` with `tomlkit` for the whole project also, as `tomlkit` is superior like I outlined [here](https://github.com/relekang/python-semantic-release/issues/275#issuecomment-743980593). But this is stuff for another issue/PR. 

### ATTENTION

**As of now** I do **not** plan to finish this PR, as I only quickly fixed the issue so I can test the tool to see if it meets my needs. Therefore I did not add any tests, but the current ones succeed altogether.

If someone wants to build on top of my changes you may hit me up by tagging me in the comments. I will be glad to add you as a collaborator to my fork so you can finish what I began.

Fixes #275

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -49,6 +49,11 @@ specify multiple versions:
         'docs/conf.py:version',
     ]
 
+When you specify a toml file with a version you must set the *toml-key-path* to
+the version string after the colon like so::
+
+    pyproject.toml:tool.poetry.version
+
 .. _config-version_pattern:
 
 ``version_pattern``
--- a/semantic_release/history/__init__.py
+++ b/semantic_release/history/__init__.py
@@ -2,10 +2,13 @@
 """
 import csv
 import logging
+import operator
 import re
+from functools import reduce
 from typing import List, Optional, Set
 
 import semver
+import tomlkit
 
 from ..errors import ImproperConfigurationError
 from ..helpers import LoggedFunction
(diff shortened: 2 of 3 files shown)


Cielquan:
Because another tool fits my needs better I will *not* continue this PR. But the above still applies.

Toilal:
I have written another PR implementing proper toml file support : https://github.com/relekang/python-semantic-release/pull/307

PR #294, opened 2020-12-16
Ensure GitHub content type is always defined (application/octet-stream)

Toilal:
If an error occurs while uploading files to github release, it's now logged as a warning and use raise_for_status feature from python request lib.

This fix issue when uploading files without extension to github. In this case, mimetype can't be guessed. Then, a POST request is performed without Content-Type header, but it fails with an error response from Github API.

### DIFF
--- a/semantic_release/hvcs.py
+++ b/semantic_release/hvcs.py
@@ -253,11 +253,15 @@ class Github(Base):
         """
         url = "https://uploads.github.com/repos/{owner}/{repo}/releases/{id}/assets"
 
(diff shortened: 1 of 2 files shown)


Toilal:
I think this line could be removed, but let me known : https://github.com/relekang/python-semantic-release/blob/1b386c6d78772fd09351697c83fd4e39dfae1821/semantic_release/hvcs.py#L19

Toilal:
I have rebased on master (v7.8.0).

PR #295, opened 2020-12-17
fix(logs): Fix TypeError when enabling debug logs

Toilal:
Some logger invocation were raising "TypeError: not all arguments converted during string formatting", because of invalid argument count given.

This also refactor some other logs to use f-strings.


Toilal:
I'll fix conflicts , stay tuned

Toilal:
I have rebased on master (v7.8.0).

relekang:
Thanks for all these PRs and thanks for fixing the conflicts 😄

PR #296, opened 2020-12-18
fix(netrc): Prefer using token defined in GH_TOKEN instead of .netrc …

Toilal:
…file

.netrc file will only be used when available and no GH_TOKEN environment variable is defined.

This also add a test to make sure .netrc is used properly when no GH_TOKEN is defined.

### DIFF
--- a/semantic_release/hvcs.py
+++ b/semantic_release/hvcs.py
@@ -7,6 +7,7 @@ from typing import Optional
 
 import gitlab
 import requests
+from requests.auth import AuthBase
 
 from .errors import ImproperConfigurationError
 from .helpers import LoggedFunction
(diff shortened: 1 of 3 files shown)


semantic_release/hvcs.py
```diff
@@ -59,6 +60,26 @@ def _fix_mime_types():
     mimetypes.add_type("text/markdown", ".md")
 
 
+class TokenAuth(AuthBase):
```
Toilal, line 63: Using a request `AuthBase` implementation make things easier to handle, as the internal requests netrc Auth is used only if no auth is configured on the request.

In previous implementation, `Authorization` header is set manually and auth is not set. So this header is replaced by internal netrc Auth implementation, making `.netrc` file having a higher priority than `GH_TOKEN` environment variable.

With this change, `GH_TOKEN` will always have the priority if defined, even if `.netrc` file is available on the host.

PR #297, opened 2020-12-19
fix(history): require semver >= 2.10

wyardley:
This resolves deprecation warnings, and updates this to a more 3.x compatible syntax


wyardley:
Don't _think_ this needs to be breaking, since the dependency is updated in `setup.py`?

If this is not conservative enough, feel free to close - this just resolves a deprecation warning observed when running tests.

wyardley:
```
error: can only concatenate str (not "VersionInfo") to str
```

sorry, that CI failure is likely from this. Let me take a look

PR #303, opened 2020-12-23
feat(repository): Add to settings artifact repository

benjaminrene:
- Add new config var to set repository (repository_url)
- Remove 'Pypi' word when it refers generically to an artifact repository system
- Depreciate 'PYPI_USERNAME' and 'PYPI_PASSWORD' and prefer 'REPOSITORY_USERNAME' and 'REPOSITORY_PASSWORD' env vars
- Depreciate every config key with 'pypi' and prefer repository
- Update doc in accordance with those changes


relekang (changes requested):
Please fix the test failures.

semantic_release/cli.py
```diff
@@ -259,28 +261,28 @@ def publish(**kwargs):
         # Get config options for uploads
         dist_path = config.get("dist_path")
         remove_dist = config.get("remove_dist")
-        upload_pypi = config.get("upload_to_pypi")
-        upload_to_pypi_glob_patterns = config.get("upload_to_pypi_glob_patterns")
+        upload_to_artifact_repository = config.get("upload_to_repository") and config.get("upload_to_pypi")
```
jkuball, line 264: Shouldn't this be an `or`?

danth: These default to true, using `and` means only one option needs to be disabled.

danth: This PR depreciates `upload_to_pypi` renaming it to `upload_to_repository`, so they are essentially the same option. The custom repository will be set to PyPI by default.

jkuball: Ah, I *am* overlooking something. Thanks for explaining!

rpocase:
Any update on this? I could definitely use this. I'm working on adding this to a poetry based project that uploads to a gitlab instance package repo. I'm having to detect if a new tag was created prior to running `poetry publish`.

relekang:
Closed in favor of #395

PR #308, opened 2021-01-06
feat(build): allow falsy values for build_command to disable build step

Toilal:
It may sounds strange to have no `build_command` at all, but my build is performed for 3 distincts OS (linux, macos, windows) with pyinstaller inside a first github action job, and then another job grabs generated artifacts to publish using semver. There's no build command inside PSR to perform because artifacts to publish are already produced by github action.

### DIFF
--- a/semantic_release/dist.py
+++ b/semantic_release/dist.py
@@ -13,12 +13,12 @@ def should_build():
     upload_pypi = config.get("upload_to_pypi")
     upload_release = config.get("upload_to_release")
     build_command = config.get("build_command")
-    return build_command and (upload_pypi or upload_release)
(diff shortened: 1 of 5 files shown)


semantic_release/dist.py
```diff
@@ -9,6 +9,18 @@
 logger = logging.getLogger(__name__)
 
 
+def should_build():
```
relekang, line 12: Please add tests of these in `test_dist.py`

relekang (changes requested):

semantic_release/dist.py
Toilal: Done

relekang:
Thanks 👍

PR #309, opened 2021-01-06
feat(print-version): add print-version command to output version

Toilal:
This new command can be integrated in the build process before the effective release, ie. to rename some files with the version number.

Users may invoke `VERSION=$(semantic-release print-version)` to retrieve the version that will be generated during the release before it really occurs.

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -71,7 +71,7 @@ COMMON_OPTIONS = [
         "--define",
         "-D",
         multiple=True,
-        help='setting="value", override a configuration value',
+        help='setting="value", override a configuration value.',
     ),
     overload_configuration,
 ]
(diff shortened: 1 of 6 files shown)


Toilal:
In fact, most users won't need this. 

But in my use case (related to PR #308), the build is not performed by semantic-release but by some github actions running in another job. So I need to retrieve the version number that semantic-release will generate to rename the artifacts generated by github actions, before placing them in the dist directory and running the `semantic-release publish` command.

semantic_release/cli.py
```diff
@@ -390,12 +417,12 @@ def cmd_version(**kwargs):
...
-if __name__ == "__main__":
-    #
-    # Allow options to come BEFORE commands,
-    # we simply sort them behind the command instead.
-    #
-    # This will have to be removed if there are ever global options
-    # that are not valid for a subcommand.
-    #
-    entry()
+@main.command(name="print-version", help=version.__doc__)
```
relekang, line 420:
```suggestion
@main.command(name="print-version", help=print_version.__doc__)
```

relekang (changes requested):

relekang:
Nice, just added one suggestion since it seems to be a copy-paste error?

semantic_release/cli.py
Toilal: Oops :) It's fixed now, thanks.

relekang:
Just realised that changing the location of the main will break the github action. Can you please update https://github.com/relekang/python-semantic-release/blob/master/action.sh#L17 to work with the new setup?

Toilal:
good catch, I just fixed this one, thanks.

PR #337, opened 2021-05-25
fix: pin tomlkit to 0.7.0

### LINKED ISSUE #336: PSR + Poetry: most of pyproject.toml deleted during release
## The problem

Running PSR via the Github action, on a project using Poetry, and [the release failed](https://github.com/browniebroke/netlify-builds/commit/09ec763e8576be83fe20cffb0597917b65254052) because PSR removed some fields which are required by Poetry: name and description.

Here is [the commit from PSR](https://github.com/browniebroke/netlify-builds/commit/09ec763e8576be83fe20cffb0597917b65254052), it removed most of the `pyptoject.toml` content.

## Expected behavior

Package is released without errors. It was working 27 days ago at least (last release).

## Additional context

Even more confusing, the only release of PSR we had in the meantime is 7.15.4 which seems like a pretty trivial change.

browniebroke:
Doesn't look like the proper fix, but [it solved the problem](https://github.com/browniebroke/netlify-builds/runs/2669817097?check_suite_focus=true#step:4:285) for me. 

Fixes #336

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
         "requests>=2.25,<3",
         "wheel",
         "python-gitlab>=1.10,<3",
-        "tomlkit>=0.7.0,<1.0.0",
+        "tomlkit==0.7.0",
         "dotty-dict>=1.3.0,<2",
     ],
     extras_require={


relekang:
Thanks 👍

PR #339, opened 2021-06-08
fix:update click and tomlkit

lyz-code:


setup.py
```diff
@@ -43,7 +44,7 @@ def _read_long_description():
         "requests>=2.25,<3",
         "wheel",
         "python-gitlab>=1.10,<3",
-        "tomlkit==0.7.0",
+        "tomlkit==0.7.2",
```
browniebroke, line 47: FYI this was [pined to 0.7.0](https://github.com/relekang/python-semantic-release/pull/337) as it seems it was [causing a bug ](https://github.com/relekang/python-semantic-release/issues/336) where the pyproject.toml was being wiped. 

Should probably have left a comment or even better write a test reproducing the problem... 😞

grawlinson:
Has this issue been reported upstream?

browniebroke:
So it turns out to be a problem only when the TOML file has out-of-order tables. The spec [discourages them](https://toml.io/en/v1.0.0#table) ("Defining tables out-of-order is discouraged."), but they are considered valid.

I tried to look at `tomlkit`, but it has [several](https://github.com/sdispater/tomlkit/blob/627fb1ae496a2531ec738e00c1abf6339472c1a4/tests/test_toml_document.py#L543) [tests](https://github.com/sdispater/tomlkit/blob/627fb1ae496a2531ec738e00c1abf6339472c1a4/tests/test_toml_document.py#L577) for them, and it seems to work fine on its own. However, when used with `dotty-dict` -like it's used in PSR- we see the problem from #336, so I'm not too sure to which upstream this should be reported: `tomlkit` or `dotty-dict`...

I've added [a test case](https://github.com/relekang/python-semantic-release/blob/ea9f6e28c58d70733c09ca57b13815ba86d8bbb2/tests/history/test_version.py#L250-L264) in #349, to make sure we get the right behaviour in PSR (at least).

PR #352, opened 2021-06-24
fix: use release-api for gitlab

FirstSanny:
Used the release api from python-gitlab instead of the outdated tag api.

Solves this [issue](https://github.com/relekang/python-semantic-release/issues/350)

### DIFF
--- a/semantic_release/hvcs.py
+++ b/semantic_release/hvcs.py
@@ -5,7 +5,7 @@ import mimetypes
 import os
 from typing import Optional, Union
 
-import gitlab
+from gitlab import gitlab, exceptions
 from requests import HTTPError, Session
 from requests.auth import AuthBase
 from urllib3 import Retry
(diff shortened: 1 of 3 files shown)


relekang:
Nice, can you please fix the broken tests?

FirstSanny:
> 
> 
> Nice, can you please fix the broken tests?

fixed them. but since the test is only running against mocks i am not sure how safe these tests are. 😄

relekang:
Hehe, valid point 😅 
The mocks for these integrations should probably be at the http layer.

FirstSanny:
> 
> 
> Hehe, valid point 😅
> The mocks for these integrations should probably be at the http layer.

But i guess this would be a new issue 😉 

Do need to provide anything else?

relekang:
Of course, did not mean that it should be done here. I forgot to check back after running the github action.

Thanks for fixing this 🙌

FirstSanny:
> 
> 
> Of course, did not mean that it should be done here. I forgot to check back after running the github action.
> 
> Thanks for fixing this 🙌

No problem. 😄

Issue #363, opened 2021-08-02
503 Error at changelog publish on shared domainname Gitlab


ziouf:
https://github.com/relekang/python-semantic-release/blob/83798b7bf94569014e00a5f030ac46a5d4992fff/semantic_release/hvcs.py#L363

You should use ${CI_SERVER_URL} instead of ${CI_SERVER_HOST} to build Gitlab.api_url()

You could implement like this:

```python
    @staticmethod
    def api_url() -> str:
        """Gitlab api_url property
        :return: The Gitlab instance API url
        """
        return config.get("hvcs_domain", os.environ.get("CI_SERVER_URL", "https://gitlab.com"))
```

This may fix the issue HTTP 503 issue at release posting phase when Gitlab server is behind a shared domainname with a url like this : https://<shared_domain>/gitlab

error : 
```
Posting changelog to HVCS
error: 503: <html><body><b>Http/1.1 Service Unavailable</b></body> </html>
ERROR: Job failed: exit code 1
```

ziouf:
After some quick investigation, the use of CI_SERVER_HOST variable is not the root cause.

But the issue remains

ziouf:
Looks related to https://github.com/relekang/python-semantic-release/issues/350#issuecomment-861024217

seba-ban:
The same goes for [changelog](https://github.com/relekang/python-semantic-release/blob/master/semantic_release/changelog/changelog.py#L20).
Assuming https will produce incorrect links on http servers. As @ziouf mentioned, CI_SERVER_URL would be a better choice here. There's also a CI_SERVER_PROTOCOL variable in GitLab CI jobs.

ziouf:
https://github.com/relekang/python-semantic-release/blob/83798b7bf94569014e00a5f030ac46a5d4992fff/semantic_release/vcs_helpers.py#L97-L110

I think it's better to use Gitlab-CI environement variables if available to get owner and project name values.

CI_PROJECT_NAMESPACE
CI_PROJECT_NAME

codejedi365:
Resolved with #364

PR #364, opened 2021-08-03
Prefer Github/Gitlab environment variables to get owner/project

ziouf:
related to #363

### DIFF
--- a/semantic_release/vcs_helpers.py
+++ b/semantic_release/vcs_helpers.py
@@ -20,8 +20,7 @@ from .settings import config
 try:
     repo = Repo(".", search_parent_directories=True)
 except InvalidGitRepositoryError:
-    # repo = None
-    raise GitError("Not in a valid git repository")
+    repo = None
 
 logger = logging.getLogger(__name__)
 
--- a/semantic_release/vcs_helpers.py
+++ b/semantic_release/vcs_helpers.py
@@ -9,7 +9,8 @@ from pathlib import Path, PurePath
 from typing import Optional, Tuple
 from urllib.parse import urlsplit
 
-from git import GitCommandError, InvalidGitRepositoryError, Repo, TagObject
+from git import GitCommandError, InvalidGitRepositoryError, Repo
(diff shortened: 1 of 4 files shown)


relekang (changes requested):
Please fix the test by mocking the environment variable and add a test for using the environment variable.

ziouf:
> 
> 
> Please fix the test by mocking the environment variable and add a test for using the environment variable.

Added some tests and updated upstream

ziouf:
Could you review again this pull request ?
Thank you

semantic_release/vcs_helpers.py
```diff
@@ -19,7 +20,8 @@
 try:
     repo = Repo(".", search_parent_directories=True)
 except InvalidGitRepositoryError:
-    repo = None
+    # repo = None
+    raise GitError("Not in a valid git repository")
```
relekang, line 24: Why do we need to raise an error here?

ziouf: does it make sense to run python-semantic-release outside of a valid git repository ?
I agree it could be managed diffrently. 
Generally I prefer an error message followed by exit rather noop.

relekang: This code runs on import as a part of a command and thus buy passes the error handling.

ziouf: @relekang mypy is failing because of your initial implement
 
https://github.com/relekang/python-semantic-release/runs/4270162688?check_suite_focus=true#step:5:8

Maybe you need to change error handling implement. Or not initialize repo variable in global scope ...

fleXible: instead of assigning None a simple pass will do, because it will set repo to None implicitly and mypy won't complain. And it's safe to do that, because the `check_repo' wrapper checks for repo for None and raises an exception.
Instead of using a global variable, it would be better to just wrap the whole thing in a class.

relekang:
Thanks, merged it locally to fix the conflicts

PR #371, opened 2021-08-07
doc: add undocumented changelog_section value

FirefoxMetzger:


docs/configuration.rst
```diff
@@ -251,7 +251,7 @@ They will be displayed in the order they are given.
 
 The available options depend on the commit parser used.
 
-Default: `feature, fix, breaking, documentation, performance` plus all
+Default: `feature, fix, breaking, documentation, performance, Other` plus all
```
relekang, line 254: Why is Other not other like the rest of them?

FirefoxMetzger: I agree that it should be "other" like the rest; however "Other" is a section that actually exists in the code, Hence I added it here for completeness.

danth: "Other" is from the Gitmoji parser whereas the rest of the sections are from the Angular parser.

FirefoxMetzger: I see, @danth , I didn't know that. In that case, it may make sense to not document it here (like the other gitmoji sections) and close this PR.

... or document the declared gitmoji sections, too, since they, too, are recognized when leaving this variable untouched in downstream config.

PR #372, opened 2021-08-10
docs(parser): documentation for scipy-parser

FirefoxMetzger:
I struggle a bit navigating these docs, so I am probably committing a blunder here. I added a paragraph to the list of parsers and then referred below for details where I `..automodule:`ed the docstring that I wrote for the module.

(I then tamed the docstring to fit a little better with the existing style of the docs.)


FirefoxMetzger:
Also @relekang , what do you think about turning on PR docs for this repo in RTD? It is as quick as ticking a box on the RTD website and then we can see how the docs render on RTD before committing them.

relekang:
Thanks. PR builds for RTD looks nice, I enabled it now.

Issue #334, opened 2021-04-23
Option to create changelog when bumping version


czee:
## Description

Currently you can only generate or update the changelog when running publish, but it would be useful in some instances to only update the changelog and bump the version without publishing. As it exists now the `changelog` command only prints to stdout.

laercio-barbosa:
@czee 

Could we assume that this request will only provide local commit? I mean, it wouldn't publish to remote HVCS, but commit locally is fine? If so, we could have this implemented in this way:

1.  `semantic-release changelog --post` would bump new version, generate changelog and push to **remote** repository.
2.  `semantic-release changelog --noop` assume the current behavior, that is, only prints changelog to standard output
3.  `semantic-release changelog` would bump new version, generate changelog and commit to **local** repository.

relekang:
The changelog command should not bump the version, if you want to bump the version without publishing use the version command. We can add a `--post` flag to changelog that is passed on to the changelog function.

czee:
Sorry for the late reply. I think everything you mentioned, @laercio-barbosa is optimal taking into account what @relekang said. So, no changelog command that bumps version, but is still able to create a changelog when other commands are run.

laercio-barbosa:
So, if I understood...

semantic-release changelog --post --> generate changelog and push it to remote repository (it is current flag behavior).
semantic-release changelog --noop --> assume the current behavior without flags, that is, only prints changelog to standard output
semantic-release changelog --> generate changelog and commit to local repository.


@relekang 
You mentioned about to add `--post` flag to changelog command, but this is already in place, right?

czee:
Seems good to me

relekang:
>You mentioned about to add --post flag to changelog command, but this is already in place, right?

What I meant is add the click decorator to [`cmd_changelog`](https://github.com/relekang/python-semantic-release/blob/master/semantic_release/cli.py#L403-L414) so it gets passed to `changelog`.

>semantic-release changelog --post --> generate changelog and push it to remote repository (it is current flag behavior).
semantic-release changelog --noop --> assume the current behavior without flags, that is, only prints changelog to standard output
semantic-release changelog --> generate changelog and commit to local repository.

Sounds good to me, changing the last one there will be a breaking change tho.

raddessi:
@relekang what are your thoughts on https://github.com/relekang/python-semantic-release/pull/390?

codejedi365:
I believe everything requested here is doable with the latest version.

PR #390, opened 2021-10-07
feat: add ability to update the changelog when bumping versions

dsarmiento:
Possible implementation for #334 
Added a `--changelog` flag to the `version` command to also update the changelog file.


raddessi:
I love this. Is there a blocker?

PR #394, opened 2021-10-29
feat(parser_angular): allow customization in parser

charlesthomas:
- `angular_parser_allowed_types` controls allowed types
  - defaults stay the same: build, chore, ci, docs, feat, fix, perf, style,
    refactor, test
- `angular_parser_default_level_bump` controls the default level to bump the
  version by
  - default stays at 0
- `angular_parser_minor_types` controls which types trigger a minor version
  bump
  - default stays at only 'feat'
- `angular_parser_patch_types` controls which types trigger a patch version
  - default stays at 'fix' or 'perf'

### DIFF
--- a/semantic_release/defaults.cfg
+++ b/semantic_release/defaults.cfg
@@ -1,4 +1,8 @@
 [semantic_release]
+parser_angular_allowed_types=build,chore,ci,docs,feat,fix,perf,style,refactor,test
+parser_angular_default_level_bump=0
+parser_angular_minor_types=feat
+parser_angular_patch_types=fix,perf
 branch=master
 build_command=python setup.py sdist bdist_wheel
 changelog_components=semantic_release.changelog.changelog_headers
--- a/semantic_release/defaults.cfg
+++ b/semantic_release/defaults.cfg
@@ -1,6 +1,7 @@
 [semantic_release]
 parser_angular_allowed_types=build,chore,ci,docs,feat,fix,perf,style,refactor,test
-parser_angular_default_level_bump=0
+parser_angular_default_level_bump=no-release
+# valid options: no-release, patch, minor, major
 parser_angular_minor_types=feat
 parser_angular_patch_types=fix,perf
 branch=master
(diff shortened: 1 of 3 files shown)


charlesthomas:
i added tests, but could only get one of them to work. i assume it has something to do with the way i'm doing the mocking, but i couldn't figure it out.

if i actually change the `defaults.cfg` file then the tests pass, but when i leave them as what they should be and try to mock them the tests fail

tests/parsers/test_angular.py
```diff
@@ -97,3 +99,33 @@ def test_parser_return_footer_from_commit_message():
...
+def test_parser_custom_allowed_types():
+    assert angular_parser("custom: ...")[0] == 0
+    assert angular_parser("custom(parser): ...")[1] == "custom"
+    pytest.raises(UnknownCommitMessageStyleError, angular_parser, "feat(parser): ...")
+
+
+@mock.patch("semantic_release.history.parser_angular.config.get",
+            wrapped_config_get(angular_parser_minor_types='docs'))
+def test_parser_custom_minor_types():
+    assert angular_parser("docs: write some docs")[0] == 2
```
esatterwhite, line 125: Don't you need to also set the allowed types

charlesthomas: `docs` and `test` are supported by default, so for those tests they should only need what's there. i did try adding it for docs, just to confirm, and the test still failed the same way.

for the custom one, i added `custom` and removed `feat`, but i didn't change the default bump level. if `custom` was allowed it should return 0 and not throw an exception (but it is throwing an exception)

for `feat` since it was removed, it _should_ throw an exception, but if you comment out the assertions about `custom`, the test fails because it _doesn't_ throw an exception

semantic_release/defaults.cfg
```diff
@@ -1,4 +1,8 @@
 [semantic_release]
+angular_parser_allowed_types=build,chore,ci,docs,feat,fix,perf,style,refactor,test
```
relekang, line 2: Can we rename these to `parser_angular_*`?

relekang (changes requested):

relekang:
I merged this locally and fixed the tests, thanks for the contribution 😄

PR #395, opened 2021-11-10
feat(repository): Add to settings artifact repository

fleXible:
I'm reopening this PR, because #303 seems to be dead. I just fixed the tests, updated the docs a bit and resolved all merge conflicts to current master.

I use GitLab myself and am looking for an integrated workflow between it's package repositories, release feature and pipeline automation.

@relekang Maybe you can have a look at it. Feedback is welcome


semantic_release/repository.py
```diff
@@ -0,0 +1,62 @@
...
+from semantic_release.settings import config
+
+from .helpers import LoggedFunction
+
+logger = logging.getLogger(__name__)
+
+
+class Repository(object):
+
+    def __init__(self, username: str = None, password: str = None) -> None:
```
relekang, line 20: Make these not optional, having the default spread between this and the get_repository method can be confusing.

semantic_release/repository.py
```diff
@@ -0,0 +1,62 @@
...
+from semantic_release import ImproperConfigurationError
+
+from semantic_release.settings import config
+
+from .helpers import LoggedFunction
+
+logger = logging.getLogger(__name__)
+
+
+class Repository(object):
```
relekang, line 18: What is the benefit of turning the function that was here into a class?

relekang (changes requested):

semantic_release/repository.py
fleXible: As stated in the PR, I'm not the original author of most of the code, so I can just take a guess. The original author is Benjamin René from #303 
My contributions are documentation, cleanups & fixing the failed tests.

relekang: Oh, sorry 🙈

fleXible: No worries, I made some progress last weekend but still needs some cleaning up before I can put it up. At least it's working now, but still far from perfect.
Still I just pushed it to get some feedback.

semantic_release/cli.py
```diff
@@ -344,6 +332,31 @@ def publish(**kwargs):
     # else: Since version shows a message on failure, we do not need to print another.
 
 
+def repo(**kwargs: Dict[str, Any]):
```
fleXible, line 335: Added that to quickly be able to do integration testing.

fleXible:
Okay, I think it's done. Tests should all be passing, at least they do locally on my system.

fleXible:
Obviously "should be passing" can be quite dramatically different from "are passing".
I had another go at it, now it shall be better. I couldn't fix the 2nd commitlint error, as it refers to a commit that's now burried deep in the history so rewording landed me in rebase-hell. And secondly it's not even a commit I authored.

relekang:
Thanks for taking the time to finish this 😄

fleXible:
> Thanks for taking the time to finish this 😄

To be honest, it was not easy, but I learned a lot in the process. 
And it's always good to see some code working the way I imagined it in the beginning!

Btw. I've found the "common options" not beeing common at all. Like some commands do have no-op handling, while others don't, or only partially.

relekang:
Yeah, I think this codebase at the moment is suffering a bit from a lot of small fixes without a common direction breaking things like the common options. I think they used to work at some point 🤷

fleXible:
> Yeah, I think this codebase at the moment is suffering a bit from a lot of small fixes without a common direction breaking things like the common options. I think they used to work at some point 🤷

I find the whole "common options" quite confusing. A big part of that comes from the fact, that   "--major", "--minor", "--patch" and "--retry" are separate flags, but they exclude each other.
Also the effect of "--post" with commands other than "changelog" is unclear, probably has no effect. "publish" will post the changelog by default with no way to disable it.

PR #397, opened 2021-11-17
feat: support github enterprise server

### LINKED ISSUE #378: support github enterprise server (GHE) in CI
## Description

It looks like `hvcs_domain` configuration seems to almost solve this but the same value is being used for the api and domain urls which is a problem.

https://github.com/relekang/python-semantic-release/blob/83798b7bf94569014e00a5f030ac46a5d4992fff/semantic_release/hvcs.py#L97-L104

https://github.com/relekang/python-semantic-release/blob/83798b7bf94569014e00a5f030ac46a5d4992fff/semantic_release/hvcs.py#L107-L116

Instead, it would be better to have a separate config value for api domain to use with custom domains.

Also, to provide built-in support for Github CI and Github enterprise server CI, it would be best to get these values from environment variables (without having to set `hvcs_domain` for enterprise CIs).

GITHUB_SERVER_URL | Returns the URL of the GitHub server. For example: https://github.com.
-- | --
GITHUB_API_URL | Returns the API URL. For example: https://api.github.com.

[Github CI Docs ref](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables)

## Use cases

- Can be used in GHE CI and GH CI

## Possible implementation

- To support custom domains, a new configuration variable called `hvcs_api_domain` can be used and then the `api_url` function can be changed to: 

```python
class Github(Base):
    # rest of the code

    @staticmethod
    def api_url() -> str:
        """Github api_url property
        :return: The Github API URL
        """
        # not necessarily prefixed with api in the case of a custom domain, so
        # can't just default DEFAULT_DOMAIN to github.com
        hvcs_api_domain = config.get("hvcs_api_domain")
        hostname = hvcs_api_domain if hvcs_api_domain else "api." + Github.DEFAULT_DOMAIN
        return f"https://{hostname}"

    # rest of the code
```

- To support Github and GHE CI, the `domain` and `api_url` methods in Github class can be modified as:

```python
class Github(Base):
    """Github helper class"""

    DEFAULT_DOMAIN = "github.com"
    _fix_mime_types()

    @staticmethod
    def domain() -> str:
        """Github domain property
        :return: The Github domain
        """
        hvcs_domain = config.get("hvcs_domain", os.getenv("GITHUB_SERVER_URL", "").replace("https://", ""))
        domain = hvcs_domain if hvcs_domain else Github.DEFAULT_DOMAIN
        return domain

    @staticmethod
    def api_url() -> str:
        """Github api_url property
        :return: The Github API URL
        """
        # not necessarily prefixed with api in the case of a custom domain, so
        # can't just default DEFAULT_DOMAIN to github.com
        hvcs_api_domain = config.get("hvcs_api_domain", os.getenv("GITHUB_API_URL", "").replace("https://", ""))
        hostname = hvcs_api_domain if hvcs_api_domain else "api." + Github.DEFAULT_DOMAIN
        return f"https://{hostname}"

    ... # rest of the code
```

ukalwa:
fixes #378

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -468,6 +468,19 @@ The name of your hvcs. Currently only ``github`` and ``gitlab`` are supported.
 
 Default: `github`
 
+.. _config-hvcs_domain:
+
+``hvcs_domain``
+---------------
+The domain url (without https://) of your custom vcs server.
+
+.. _config-hvcs_api_domain:
+
+``hvcs_api_domain``
+-------------------
+The api url (without https://) of your custom vcs server.
+
+
 .. _config-check_build_status:
 
 ``check_build_status``
--- a/semantic_release/hvcs.py
+++ b/semantic_release/hvcs.py
@@ -101,7 +101,10 @@ class Github(Base):
 
         :return: The Github domain
         """
-        hvcs_domain = config.get("hvcs_domain")
+        # ref: https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
+        hvcs_domain = config.get(
+            "hvcs_domain", os.getenv("GITHUB_SERVER_URL", "").replace("https://", "")
+        )
         domain = hvcs_domain if hvcs_domain else Github.DEFAULT_DOMAIN
         return domain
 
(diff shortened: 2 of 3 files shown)


relekang:
Please rebase and check that it works with latest changes.

relekang:
Thanks 😄

Issue #404, opened 2022-01-17
Proposals improve


mom1:
## Description

Hi.
First of all, I want to thank you for your work.
I apologize in advance my English is very bad 😄 

I had a few ideas for improvement and wanted to ask you if you would like to include some of them in your repository?

### [python-semantic-release](https://github.com/relekang/python-semantic-release/pull/405)

The biggest limitation for me at the moment is the `semantic_release.history.logs.generate_changelog` method.
It returns not enough information about the commit, I would like to offer you a refinement with the possibility of overriding it in the `commit_analyzer` setting.

### [python-semantic-release-changelog-template](https://github.com/mom1/python-semantic-release/commits/python-semantic-release-changelog-template)

In this plugin, I would like to allow the user to generate a changelog using `mustache` templates. To do this, use our `commit_analyzer` to fill the list of commits with something like this [data](https://semantic-release.gitbook.io/semantic-release/developer-guide/js-api#commits) and put them to template.

### [python-semantic-release-gitmoji](https://github.com/mom1/python-semantic-release/commits/python-semantic-release-gitmoji)

A slightly different implementation of the emoji data from [gitmoji](https://github.com/carloscuesta/gitmoji)
A bit like [this](https://github.com/momocow/semantic-release-gitmoji)

### [python-semantic-release-changelog-processing](https://github.com/mom1/python-semantic-release/tree/python-semantic-release-changelog-processing)

In this plugin, I would like to allow the user to format subject and body commit.
For example:
```ini
# settings
changelog_body_processors=ident
changelog_subject_processors=strip,final_dot,add_pr_link_github
```
`strip,final_dot,add_pr_link_github` - each part can be
- Name function from `processors` module
- Importable string to custom processor
```
# raw commit message
feat: some (#15) subject

- some cool things
- some cool 
   things multiline things
```

```
# processed commit message
Some ([#15](https://github.com/owner/repo_name/issues/15)) subject.
  - some cool things
  - some cool 
     things multiline things
```

### [python-semantic-release-merge-request](https://github.com/mom1/python-semantic-release/commits/python-semantic-release-merge-request)

I doubt that this can be done with a separate plugin.

`semantic_release merge_request from to`
Will create merge request in gitlab with changelog 
from template and commits between `from` and `to`

`semantic_release merge_request feature/issue-22 develop`
`semantic_release merge_request 456987 develop`
`semantic_release merge_request v1.1.0 654987`

### python-semantic-release-pull-request

[...]

PR #405, opened 2022-01-17
feat(history): Add to settings commit_analyzer

mom1:
Add to settings commit_analyzer  (#404)

- default value `semantic_release.history.logs.generate_changelog`
- function for get `semantic_release.settings.current_commit_analyzer`


relekang:
It is a bit unclear to me what analyzer is vs parser which is already configurable?

PR #408, opened 2022-01-21
feat: include additional file changes in release commits

MattPColeman:
It may be useful to include additional file changes within the automated release commits (e.g. generated docs, diagrams, ...)

I suggest that we allow the user to specify a command to be run pre-commit (`pre_commit_command`) and a list of files they would like to include in that commit (`include_additional_files`).

Example usage:
```
[semantic_release]
...
pre_commit_command=make generate-diagrams
include_additional_files=img/d1.png,img/d2.png
```

Feedback welcome! I'll cleanup/document/test more if @relekang agrees with the principle


relekang:
Nice, please add docs for the options and then this will be good 😄

MattPColeman:
> Nice, please add docs for the options and then this will be good 😄

👍  Added

relekang:
Thanks 😄

PR #417, opened 2022-02-22
chore(dependencies): extend allowed version range for python-gitlab

deveaud-m:
[python-gitlab](https://github.com/python-gitlab/python-gitlab/releases) released version 3.0.0 last month

is there any reason for pining the upper version of this dependency? I can restrict it to <4 if this has proven to be useful in this project in the past

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
         "twine>=3,<4",
         "requests>=2.25,<3",
         "wheel",
-        "python-gitlab>=1.10,<3",
+        "python-gitlab>=2",
         # See https://github.com/relekang/python-semantic-release/issues/336
         # and https://github.com/relekang/python-semantic-release/pull/337
         # for why tomlkit is pinned


danth:
I think it was pinned to < 3  as a major version bump (from 2 to 3) indicates breaking changes. If everything still works with v3, bumping the pin to the next major version (< 4) is the right way to proceed.

deveaud-m:
> I think it was pinned to < 3 as a major version bump (from 2 to 3) indicates breaking changes. If everything still works with v3, bumping the pin to the next major version (< 4) is the right way to proceed.

ok, I bumped the pin to <4 in a new commit

I am quite new to mypy and trying to understand how to annotate the type of object returned by the Gitlab API call to avoid the error we see in the Github action: https://github.com/relekang/python-semantic-release/runs/5290758086?check_suite_focus=true

@danth how would you handle it?

deveaud-m:
After reading about mypy errors and looking at the `python-gitlab` code, I couldn't find a solution to the type error thrown by objects returned by the Gitlab API wrapper. Even if the `__getattr__`  class method for objects of type `RESTObject` properly returns the type *Any*, mypy still complains.

See also: https://mypy.readthedocs.io/en/stable/common_issues.html#spurious-errors-and-locally-silencing-the-checker

I decided to ignore the type error in the five related lines, at least it should help the Github Actions succeed.

deveaud-m:
@danth did you had some time to look at that change?

danth (approved):
Looks good to me, sorry for the delay

PR #419, opened 2022-02-22
Allow using ssh-key to push version while using token to publish to hvcs

sp-ricard-valverde:
# Scenario

In GH Actions I want to use a ssh Deploy Key to push changes to a _protected_ branch because not every user's GITHUB_TOKEN is allowed to push to _protected_ branches but Deploy Keys are.

# Problematic

If I specify a `GH_TOKEN` such as `GITHUB_TOKEN` in a GitHub Action with enough permission to publish a GitHub Release, python-semantic-releases will also use that user's token to push version and tag changes to the repository.
The user running the action though might not have enough permission to push to a _protected_ branch even if grating some degree of permission to its `GITHUB_TOKEN`. Only Deploy Keys or Admin user's `GITHUB_TOKEN`(or Personal Access Tokens) can.

# Solution

Add an option that will allow to use the token to submit GitHub Releases while will avoid using that token for pushing changes to the repository, relying on the underlying git authentication mechanism. In the case of GitHub Actions it can be combined with [ssh-agent action](https://github.com/webfactory/ssh-agent) to allow using Deploy Keys for this purpose.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -538,3 +538,14 @@ Variable          Contents
 ================  ========
 
 Default: ``v{version}``
+
+.. _config-ignore_token_for_push:
+
+``ignore_token_for_push``
+-------------------------
+Do not use the default auth token to push changes to the repository. Use the system
+configured method.
+This is useful if the auth token does not have permission to push, but the system method
+(an ssh deploy key for instance) does.
+
+Default: `false`
--- a/semantic_release/defaults.cfg
+++ b/semantic_release/defaults.cfg
@@ -23,6 +23,7 @@ gitlab_token_var=GL_TOKEN
 gitea_token_var=GITEA_TOKEN
 
 hvcs=github
+ignore_token_for_push=false
 major_emoji=:boom:
 major_on_zero=true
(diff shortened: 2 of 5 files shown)


sp-ricard-valverde:
Anyone ?

danth (changes requested):
Please add the new option to the documentation in `docs/configuration.rst`.

Other than that, this looks good to me.

danth (approved):

danth:
Looks like some tests need to be fixed.

sp-ricard-valverde:
I'll look into it, thanks @danth

sp-ricard-valverde:
Sorry for the late fix @danth. I've figured out the problem which had to do with the GITHUB_ACTOR env var set by the GH runners altering the test outcome(thus why locally the tests were ok)

danth:
One of the tests is still broken 🤔 Perhaps there is still some interference from the test environment.

sp-ricard-valverde:
Hope this one works. I explicitly added the GITHUB_ACTOR env var to my test execution to emulate the runner behaviour

PR #432, opened 2022-03-30
fix(prerelase): pass prerelease option to get_current_version

anehx:
The `get_current_version` function accepts a `prerelease` argument which was never passed.

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -98,7 +98,7 @@ def print_version(*, current=False, force_level=None, prerelease=False, **kwargs
     Print the current or new version to standard output.
     """
     try:
-        current_version = get_current_version()
+        current_version = get_current_version(prerelease)
     except GitError as e:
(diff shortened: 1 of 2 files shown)


anehx:
The test for this never failed because it was mocked correctly: 

https://github.com/relekang/python-semantic-release/blob/85158798ca438c1dafc84036d13c2988c934f02f/tests/test_cli.py#L357-L359

However, `get_current_version` could never have returned this value because the `prerelease_bool` argument was never passed which lead to ignoring **all** prereleases:

https://github.com/relekang/python-semantic-release/blob/85158798ca438c1dafc84036d13c2988c934f02f/semantic_release/history/__init__.py#L230

danth (changes requested):
Please fix the failing tests.

anehx:
@danth the tests should pass now..

danth (approved):

PR #433, opened 2022-03-31
doc: fix semantic_release config

donhui:


danth (commented):
Looks like you've changed this example from `setup.cfg` format to `pyproject.toml` format. Perhaps we should include both?

relekang:
Fixed in [2a4b8af](https://github.com/relekang/python-semantic-release/commit/2a4b8af1c2893a769c02476bb92f760c8522bd7a)

PR #444, opened 2022-05-06
fix: get_current_release_version incorrectly parses commit message with version

ngshiheng:
**Context:**
- Related to https://github.com/relekang/python-semantic-release/issues/442
- The most recent PR that was merged https://github.com/relekang/python-semantic-release/pull/435 seems to be causing this issue
- From my observation and tests, `get_current_release_version_by_commits` seems to be parsing version incorrectly from commit messages
- For example, this issue occurs when the commit messages that contain versioning are parsed, e.g. "fix(deps): update dependency pylint to 2.13.8" -> `2.13.8`
- More examples [here](https://pythex.org/?regex=(%5Cd%2B%5C.%5Cd%2B%5C.%5Cd%2B(%3F!-beta%5C.%5Cd%2B))&test_string=chore(deps)%3A%20update%20dependency%20pylint%20to%20v2.13.8%0Afix(deps)%3A%20update%20dependency%20sqlalchemy%20to%201.4.36%0A20.10.199111%0A1.4.1%0Av1.4.2%0A0.3.5-beta.0%0A1.41.4%0Atry%20to%20update%201.6.7%20to%201.61.1%20later&ignorecase=0&multiline=1&dotall=0&verbose=0)

**Main changes:**
- `get_commit_log` now yields an additional `commit.author` info to the returned tuple
- `get_current_release_version_by_commits` now checks that `commit_author.name == "github-actions" and commit_author.email == "action@github.com"` before returning version that matches regex in the commit message

**Description:**
- The main idea behind this change is that there is not really a guaranteed way for us to parse versions correctly from users' commit messages as users can write commit messages however they want
- This fix checks for the commit author to cope with the issue above

**How has this been tested:**
- Ran `tox` locally and passes all test cases
- Tried running `../python-semantic-release/semantic_release/.venv/bin/python -m semantic_release publish -v DEBUG --noop` locally on one of my repository that uses and failed `python-semantic-release` action recently

Let me know if this works! I'm more than happy to work together to fix this!

### DIFF
--- a/semantic_release/history/__init__.py
+++ b/semantic_release/history/__init__.py
@@ -324,7 +324,7 @@ def get_previous_version(version: str) -> Optional[str]:
     :return: A string with the previous version number.
     """
     found_version = False
-    for commit_hash, commit_message in get_commit_log():
+    for commit_hash, commit_message, _ in get_commit_log():
         logger.debug(f"Checking commit {commit_hash}")
         if version in commit_message:
             found_version = True
@@ -349,7 +349,7 @@ def get_previous_release_version(version: str) -> Optional[str]:
     :return: A string with the previous version number.
     """
     found_version = False
-    for commit_hash, commit_message in get_commit_log():
+    for commit_hash, commit_message, _ in get_commit_log():
         logger.debug(f"Checking commit {commit_hash}")
         if version in commit_message:
             found_version = True
@@ -372,10 +372,10 @@ def get_current_release_version_by_commits() -> str:
 
     :return: A string with the current version number.
     """
-    for commit_hash, commit_message in get_commit_log():
+    for commit_hash, commit_message, commit_author in get_commit_log():
         logger.debug(f"Checking commit {commit_hash}")
         match = re.search(rf"{release_version_pattern}", commit_message)
-        if match:
+        if match and commit_author.name == "github-actions" and commit_author.email == "action@github.com":
             logger.debug(f"Version matches regex {commit_message}")
             return match.group(1).strip()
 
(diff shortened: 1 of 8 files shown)


ngshiheng:
I'm not sure why mypy failed here, any ideas?

danth (changes requested):
When Python semantic release is not running under GitHub Actions, the commit author will be different. I don't see any obvious way of knowing what the author will be in these cases.

https://github.com/relekang/python-semantic-release/blob/94c94942561f85f48433c95fd3467e03e0893ab4/action.sh#L15-L17

We could parse the local Git configuration, however, some people might be using their own account details for Python semantic release commits. It's also possible for the commit author to be changed at any point in the project's history.

Currently, these changes will stop versions from being detectable in a non-Actions setting.

jacksbox:
first: sorry for introducing this issue. I wrote about this issue in context of the change-log here already https://github.com/relekang/python-semantic-release/issues/424 but somehow missed to to think about it in terms of the prerelease implementation. 

I see two options for quick solutions: 
1. rollback the prerelease changes until we have a better solution
2. add an unchangeable identifier to the commit message, e.g. always let the commit subject end with `@python-semantic-release` and use this to verify that we have the correct commit...

For the future, let's discuss here https://github.com/relekang/python-semantic-release/issues/424 if perhaps having only tags as version sources could make this package more robust and make changes / addons easier....

ngshiheng:
@danth makes sense. I didn't consider the fact that users could be using python-semantic-release without GitHub Actions. this PR won't work then. i'll go ahead and close this PR

@jacksbox no worries! i'd go with quick solution 1 as i'd expect this bug to break lots of repositories' CI. however, take my words with a grain of salt; i'm sure the maintainers know a lot better

I really appreciate you guys putting effort into this library, thank you!

Issue #456, opened 2022-06-15
With HVCS_DOMAIN set semantic-release publish still tries to upload artifacts to github.com


butler54:
## The problem
I am currently using `python-semantic-release` for a documentation based project (e.g. not python code).

```
[semantic_release]
hvcs = github
hvcs_domain = github.aprivate-domain.com
version_pattern =mkdocs.yml:ops_version: (\d+\.\d+\.\d+)
branch = main
upload_to_pypi = false
version_source = commit
build_command=false
```
Prior to semantic_release running files (pdfs) are dumped into `./dist` to be uploaded to the github enterprise server (3.4.4).

The release is created, a release is created on github enterprise server (including the 'source code'), however, when uploading the binary blob (pdf) an error such as this occurs

```
debug: upload_asset(<class 'semantic_release.hvcs.Github'>, 'github-org', 'github-repo', 4, 'dist/output.pdf')
warning: Asset upload dist/ioutput.pdf on Github has failed: 401 Client Error: Unauthorized for url: https://uploads.github.com/repos/github-org/github-repo/releases/4/assets?name=output.pdf
```
E.g. it's pointing to `github.com` rather than `github.aprivate-domain.com`.

## Expected behavior
Based on [this](https://docs.github.com/es/enterprise-server@3.5/rest/releases/assets#upload-a-release-asset) it should be possible to upload these assets to GH enterprise / enterprise server environments.

## Additional context

Feel free to shoot me a PR to test with, if required.

PR #458, opened 2022-06-19
fix: Ensure that assets can be uploaded successfully on custom github servers

butler54:
Allows custom github servers to have assets uploaded and compatible with github.com as well.,

Note: `hvcs_api_domain` is really `hvcs_api_root` for github enterprise servers, as the API root would typically be `https://my_custo_github.com/api/v3`.

Closes: #456

### DIFF
--- a/semantic_release/hvcs.py
+++ b/semantic_release/hvcs.py
@@ -2,6 +2,7 @@
 """
 import logging
 import mimetypes
+from optparse import Option
 import os
 from typing import Any, Optional, Union, cast
 from urllib.parse import urlsplit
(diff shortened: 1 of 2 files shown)


danth (commented):
Looks like you need to update the tests too.

butler54:
> Looks like you need to update the tests too.

Will do within the next week or so (+- day job commitments)

butler54:
> > Looks like you need to update the tests too.
> 
> Will do within the next week or so (+- day job commitments)

Okay the tests now pass.

danth (approved):

PR #466, opened 2022-07-11
fix: prerelease tag in version patterns

d-ryzhikov:
Delay construction of version and release patterns until runtime.
This will allow to use non-default prerelease tag.

### DIFF
--- a/semantic_release/history/__init__.py
+++ b/semantic_release/history/__init__.py
@@ -25,12 +25,18 @@ from .parser_tag import parse_commit_message as tag_parser  # noqa isort:skip
 logger = logging.getLogger(__name__)
 
 
-prerelease_pattern = f"-{config.get('prerelease_tag')}\.\d+"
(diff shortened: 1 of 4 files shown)


danth (commented):
What is the benefit of this vs what we had previously?

saseith:
> What is the benefit of this vs what we had previously?

We actually had this exact issue in one of our uses.  As the code currently stands, it is not possible to override the pre-release tag using cli flags and only works through changing it in the config files.  The commit here appears to fix that issue since it does the check after __init__

d-ryzhikov:
@saseith is right. Without this change, the "release_version_regex" and "version_regex" are assigned during the module import. Thus the "prerelease_tag" value used in the patterns will be the one available during the import time, which is before the runtime and before overwriting the config with values from the CLI.

danth:
I see; will merge this now

PR #469, opened 2022-07-12
Add packaging module to requirement to setup.py

### LINKED ISSUE #468: Github Action fails to find "packaging" module
## The problem

The packaging module is not found which causes the repository upload to fail.  [Example of failed Github Action](https://github.com/swysocki/gpt-image/runs/7293848582?check_suite_focus=true#step:4:256)

## Expected behavior

The packaging module should be installed and found so that it can be used during the Github Action.

## Additional context

The dotty-dict package previously provided the `packaging` dependency. In version 1.3.1, it no longer declares `packaging` as requirement and it is no longer installed during the Github Action.  Adding packaging to the `setup.py` fixes the issue.

swysocki:
Fixes #468

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,7 @@ setup(
         "tomlkit>=0.10.0,<0.11.0",
         "dotty-dict>=1.3.0,<2",
         "dataclasses==0.8; python_version < '3.7.0'",
+        "packaging",
     ],
     extras_require={
         "test": [


madpah:
Would be great to get this merged - seems to have broken many peoples GitHub pipelines 👎 

Shout if you would like any help!

PR #472, opened 2022-07-14
fix: ignore dependencies versions bump when parsing version from commit logs

### LINKED ISSUE #442: Workflows with python-semantic-release broke on 03.05.22
## The problem

All my repo have a nightly scheduled workflow. I noticed that all which use this Github Action broke tonight. I struggle to find what the actual error is. Here is one example where the workflow broke: https://github.com/eifinger/hass-here-weather/runs/6267125839?check_suite_focus=true
I am running on the master branch so I think [the latest commit](https://github.com/relekang/python-semantic-release/commit/94c94942561f85f48433c95fd3467e03e0893ab4) broke something.

## Expected behavior

The Github Action starts, checks commits, sees there is nothing to do and exits successfully.

## Additional context

* [Workflow Config](https://github.com/eifinger/hass-here-weather/blob/main/.github/workflows/ci.yml)
* [Semantic Release Config](https://github.com/eifinger/hass-here-weather/blob/main/pyproject.toml#L77-L85)

browniebroke:
Fix #442

### DIFF
--- a/semantic_release/history/__init__.py
+++ b/semantic_release/history/__init__.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 
 prerelease_pattern = f"-{config.get('prerelease_tag')}\.\d+"
 version_pattern = f"(\d+\.\d+\.\d+({prerelease_pattern})?)"
-release_version_pattern = f"(\d+\.\d+\.\d+(?!.*{prerelease_pattern}))"
+release_version_pattern = f"v?(\d+\.\d+\.\d+(?!.*{prerelease_pattern}))"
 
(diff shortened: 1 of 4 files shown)


eifinger:
I tested it in my repo and it works: https://github.com/eifinger/here_routing/runs/7353567801?check_suite_focus=true

browniebroke:
Ok, thanks for the feedback! I didn't know about this option, I guess that explain why `re.search` is used instead of `re.match`. Seems related to a recent change marged in #466 as well. 

I'll see if I can take this into account 👍🏻

browniebroke:
Please chek out #476 for a more up-to-date fix

PR #480, opened 2022-07-25
fix: don't use commit_subject for tag pattern matching

eifinger:
When `commit_subject` is configured but no commits are made by this lib or `version_source=tag_only` is used the changes in https://github.com/relekang/python-semantic-release/pull/476 introduced an error.

Now the `commit_subject` is only used to find the version pattern in commit messages and not in files or tags.


danth (approved):
Looks good to me!

PR #488, opened 2022-08-13
fix: search for release commit with commit_subject

d-ryzhikov:
Use release commit template settings in release commit search.


bernardcooke53:
@d-ryzhikov this looks like a good catch! Since we don't want to identify the current version with a commit that doesn't match the specified `commit_subject`, is it worth adding an explicit test that a commit in the log which _doesn't_ match the `commit_subject` template but contains the current version returns `None`?

d-ryzhikov:
@bernardcooke53 I've added a test for a case when there is a commit with a message containing current version but the subject doesn't match `commit_subject` setting. However, in this case `evaluate_version_bump` shouldn't return `None`. Instead it should not evaluate this commit as the release commit and continue the search.

bernardcooke53:
@d-ryzhikov looks good! Yes you're right about the bump, in my head I was thinking of an out-of-scope commit not triggering a bump - the way you've done it is good too :slightly_smiling_face:

d-ryzhikov:
I updated the release commit search logic in `evaluate_version_bump` and in `generate_changelog`. Now the release commit check is more strict: a commit message for a target version is constructed and only commits with exactly matching messages are considered to be target version commits.

danth (commented):
This will break if the `commit_subject` is reconfigured - but I think it's an improvement over the current behaviour.

Issue #490, opened 2022-08-22
7.31.3 keeps making new releases


zjgoodman:
## The problem

We noticed today that our CICD generated 10+ new releases

![image](https://user-images.githubusercontent.com/8799628/186010055-c7684911-bc34-4129-b6c1-03277d2bdeab.png)

After looking into it we've discovered that the problem is the newest version of SR itself. The current version of SR (7.31.3) keeps making new releases even if we don't push any new commits. Just running it on the clean master branch causes it to make a new release.
```
❯ pip list | grep python-semantic-release                                                                                                                        
python-semantic-release 7.31.3

❯ semantic-release publish --noop                                                                                                                                   
Current version: 14.0.0, Current release version: 14.0.0
warning: No operation mode. Should have bumped from 14.0.0 to 15.0.0
```

The previous version (7.31.2) works as expected
```
❯ pip list | grep python-semantic-release                                                                                                                    
python-semantic-release 7.31.2

❯ semantic-release publish --noop                                                                                                                                   
Current version: 14.0.0, Current release version: 14.0.0
No release will be made.
```

*A clear and concise description of what the bug is.*

## Expected behavior

If I run semantic release without changing any of the code in master then I won't get a release. But the actual behavior is we do get a release. This is not the case for previous versions of SR.

## Additional context
Here's the command with debug verbosity
```
❯ semantic-release publish --verbosity=DEBUG --noop  
debug: get_current_version_by_config_file()
debug: Parsing current version: path=PosixPath('src/myproject/__init__.py') pattern='__version__ *[:=] *["\\\'](\\d+\\.\\d+\\.\\d+(-dev\\.\\d+)?)["\\\']' num_matches=1
debug: Parsing current version: path=PosixPath('pyproject.toml') pattern='version *[:=] *["\\\'](\\d+\\.\\d+\\.\\d+(-dev\\.\\d+)?)["\\\']' num_matches=1
debug: Regex matched version: 14.0.0
debug: get_current_version_by_config_file -> 14.0.0
debug: get_current_release_version_by_commits()
debug: Checking commit 797c773f6be96b0224cdba6e05b1dad20b142ab7
debug: Version matches regex 14.0.0
debug: 
debug: Automatically generated by python-semantic-release
debug: get_current_release_version_by_commits -> 14.0.0
Current version: 14.0.0, Current release version: 14.0.0
debug: evaluate_version_bump('14.0.0', None)
debug: Reference v14.0.0 does not exist, considering entire history
[...]

PR #495, opened 2022-08-23
fix: account for trailing newlines in commit messages

bernardcooke53:
Resolves issue #490


bernardcooke53:
Note this is a bit brittle - but getting this out & preventing countless infinite loops in automated CI seems like a higher priority. Will need to check that commit messages are stripped for comparison. Changing in `get_commit_log` seems to break tests for changelog generation, so better to strip it at the comparison site for now imo.

The bug causing #490 is that in git logs, commit messages have a trailing newline, which isn't accounted for in the message built up from config.

Have added some extra test cases to check for this.

@d-ryzhikov - hopefully this doesn't break the testing/use case you had for implementing [your PR](https://github.com/relekang/python-semantic-release/pull/488)?
@zjgoodman - does this fix #490 for you?

danth (approved):

danth:
Going to merge this straight away so that we don't affect any more repositories with the infinite looping.

PR #502, opened 2022-09-24
feat: add setting for enforcing ascii changlelog sections

### LINKED ISSUE #498: Plain Text Changelog Section Description Setting for Emoji Parser
## Description

As a developer I really like controlling semantic versioning using emojis, however the resulting changelog is not very useful for people that are not actively involved with the project as the section descriptions are just emojis. The user would have to know that the project uses an emoji parser and what each emoji represents in the semantic versioning scheme.

## Use cases

A better experience for people reading the changelog would be to have the sections just be in plain text describing the change as patch, minor or major.

## Possible implementation

The optimal solution would be to have a setting to control whether the emoji parser should use plain text or emojis for the section descriptions. (Maybe even a combination of both)

cowofevil:
Resolves #498

Add the 'enforce_ascii_changelog_sections' setting flag for enforcing that changelog section headings will always be regular ASCII when using the Emoji parser.


danth (commented):
Code looks good to me.

Perhaps `use_textual_changelog_sections` would be a more fitting name? Mentioning ASCII specifically is an unnecessary level of detail.

cowofevil:
@danth Agreed, I think ``use_textual_changelog_sections`` is more understandable.

danth (approved):

Issue #168, opened 2020-02-14
Changelogs for multiple projects within one repository


AlphaMycelium:
As suggested [here](https://github.com/AlphaMycelium/psr-action/issues/4#issue-565453177), it would be useful to have the option to differentiate between multiple projects within the same repository when posting changelogs.

I think the most plausible solution is to pass some text on the command line which is appended before the title of the release, such as `--project "Foo"`. This would create a title of `Foo vX.X.X`. Alternatively, more customization of the ordering could be given using some sort of format string.

Obviously, if the option is omitted then the behavior would be the same as it is now.

KingDarBoja:
I forgot to mention, the generated assets on the releases tab contain the entire github repo instead of build artifacts per package. Should be good to provide extra documentation about this in some example.

Cheers!

AlphaMycelium:
Looking at that link you sent, maybe it'd be better to add more comprehensive support for monorepos instead of just the changelog titles.

wyardley:
The docs are a little thin, but there's at least some basic support for multiple projects in a repo -- https://python-semantic-release.readthedocs.io/en/latest/automatic-releases/github-actions.html#multiple-projects mentions having multiple projects setup in different directories already -- and I could see a separate changelog being generated in each directory easily enough, but I imagine this would create conflicts with the tags and GH versions?

I guess configuring a different `tag_format` in the config for each project within the directory structure would handle tags Ok - and that would presumably also handle the release name as well.

If the goal is to have a single changelog and releases, but multiple "things" internally, figuring out the "right" behavior could get pretty complicated.

codejedi365:
Still in backlog

codejedi365:
Still in backlog

aguynamedben:
One quirk I found when using python-semantic-release for a monorepo is that depending on how your CI is setup, you can't make commits with commit messages such as:

```
feat(api,ui): Adjust webpage and API calls to power the webpage
```

In my case, I had two parallel GitHub Actions "workflows" watching `master`. A single commit that touches multiple project was kicking of my workflows "API Release" and "UI Release" in parallel. The first one to get to the commit stage would work, but the second would now be out-of-sync with master so the push would fail. I didn't see any option to force push --rebase back to master.

For now I guess we'll just say "don't do that, make two separate commits - one to bump each app".

aguynamedben:
[...]

PR #509, opened 2022-10-17
feat: add option to only parse commits for current working directory

kibidev:
When running the application from a sub-directory in the VCS, the option use_only_cwd_commits will filter out commits that does not changes the current working directory, similar to running commands like `git log -- .` in a sub-directory.

Relevant, but not complete solution, for #168.

### DIFF
--- a/docs/automatic-releases/github-actions.rst
+++ b/docs/automatic-releases/github-actions.rst
@@ -104,7 +104,9 @@ multiple projects.
        repository_password: ${{ secrets.PYPI_TOKEN }}
 
 .. note::
-  The release notes posted to GitHub will not currently distinguish which
-  project they are from (see `this issue`_).
+  There is only partly support for monorepo (see `this issue`_).
+  In order to use this action, you should at least configure (in each project):
+  ``use_only_cwd_commits = true`` and a unique ``tag_format``.
+  Only ``version_source = commit`` (default) is supported.
 
 .. _this issue: https://github.com/python-semantic-release/python-semantic-release/issues/168
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -561,3 +561,22 @@ This is useful if the auth token does not have permission to push, but the syste
 (an ssh deploy key for instance) does.
 
 Default: `false`
+
+.. _config-use_only_cwd_commits:
+
+``use_only_cwd_commits``
+------------------------
+Only consider commits that are relevant for the current working directory when parsing commits.
+This can be helpful for a monorepo, where commits to one component should not trigger
+new versions for all other components as well.
+
+Default: `false`
+
+.. note::
+  Only valid if ``version_source = commit``.
+  You will probably also need to set a unique ``tag_format`` for each component.
+  Monorepo support is still only partly supported (see `this issue`_).
+
+.. _this issue: https://github.com/relekang/python-semantic-release/issues/168
+  
+  
--- a/semantic_release/defaults.cfg
+++ b/semantic_release/defaults.cfg
@@ -46,3 +46,4 @@ upload_to_pypi=true
 upload_to_release=true
 version_source=commit
 prerelease_tag=beta
+use_only_cwd_commits=false
--- a/semantic_release/settings.py
+++ b/semantic_release/settings.py
@@ -50,6 +50,7 @@ def _config_from_ini(paths):
         "upload_to_repository",
         "upload_to_release",
         "tag_commit",
+        "use_only_cwd_commits",
     }
 
     # Iterate through the sections so that default values are applied
--- a/semantic_release/vcs_helpers.py
+++ b/semantic_release/vcs_helpers.py
@@ -17,8 +17,11 @@ from .helpers import LoggedFunction
 from .settings import config
 
 _repo: Optional[Repo]
+_sub_directory = "."
 try:
     _repo = Repo(".", search_parent_directories=True)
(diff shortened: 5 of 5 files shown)


bernardcooke53:
@kibidev hey 👋 just wanted to ask a question about this - how do you anticipate it will affect the changelog generation and tagging? If you run from multiple subdirectories, won't you get multiple releases/tags, and the versions for each subproject jumping outside of what the commit semantics imply? 
E.g. in this structure:
```
repo
|---- projA  # at v1.2.0
|---- projB   # at v1.1.0
```
If you configured `version_source = "tag"`. When you release a `feat` in `projA` won't you get projA at a tag `v1.3.0`, and if you then do a `fix` in projB and rerun, with this setup you'll get a `v1.3.1` instead of `v1.1.1`?
Even if you format the project names into the tags, `projB` would jump from `projB-v1.1.0` to `projB-v1.3.1` since PSR will identify a `fix` within `projB`'s directory, but `get_latest_release_version` will return `v1.3.0` since it considers the entire repo history?

I think it would help to have some tests to demonstrate the behaviour a bit, and some more documentation about how this would impact tagging overall 🙂

kibidev:
@bernardcooke53 Hi!

> how do you anticipate it will affect the changelog generation and tagging? If you run from multiple subdirectories, won't you get multiple releases/tags, and the versions for each subproject jumping outside of what the commit semantics imply?

I have followed a setup as in
https://python-semantic-release.readthedocs.io/en/latest/automatic-releases/github-actions.html#multiple-projects
In order not to get problems with the tagging, each sub-project needs a unique `tag_format` configured. This will fix releases as well, but the "lastest release" featured will only be one of the project, whichever was generated last.

I have not tested with `version_source = "tag"`, but this PR does not change any default behavior. I'm not sure if filtering out commits for the sub directory is relevant for `version_source = "tag"` or not.

Regarding documentation, I would suggest updates the the documentation for [mult projects](https://python-semantic-release.readthedocs.io/en/latest/automatic-releases/github-actions.html#multiple-projects). Any other suggestions?

Regarding tests, I probably will not have time to add this very soon, but I will try when I have time to spare.

bernardcooke53:
@kibidev
> I have followed a setup as in
https://python-semantic-release.readthedocs.io/en/latest/automatic-releases/github-actions.html#multiple-projects
In order not to get problems with the tagging, each sub-project needs a unique tag_format configured. This will fix releases as well, but the "lastest release" featured will only be one of the project, whichever was generated last.

That helps understand the use-case - thanks! I think the scenario I outlined above is applicable in that case, so as you say adding some documentation explaining the limitations might help. Something like the note for [`commit_author`](https://python-semantic-release.readthedocs.io/en/latest/configuration.html#commit-author) explaining that `use_only_cwd_commits` doesn't provide full monorepo support with the explanation you just gave above :slightly_smiling_face:.

I do think that there's a danger when used with `version_source = "tag"` that this introduces some unexpected behaviour with respect to version increments for individual projects - while you are able to filter the commits down to just those that have objects in the tree from within a subpath, the chain of function calls when using `version_source = "tag"` ([`get_last_version`](https://github.com/kibidev/python-semantic-release/blob/d2216e1e32a9d674f1919ee93f1016eb49a6ebf9/semantic_release/vcs_helpers.py#L73#L95) and [`get_current_(release)_version_by_tag`](https://github.com/kibidev/python-semantic-release/blob/d2216e1e32a9d674f1919ee93f1016eb49a6ebf9/semantic_release/history/__init__.py#L202#L229) are the most relevant) don't filter down by tags only on such commits. For that reason, I think it would be much safer behaviour to change [this line](https://github.com/relekang/python-semantic-release/blob/d2216e1e32a9d674f1919ee93f1016eb49a6ebf9/semantic_release/vcs_helpers.py#L23) from
```python
    if config.get("use_only_cwd_commits"):
```
to
```python
    if config.get("use_only_cwd_commits") and config.get("version_source") == "commit":
```
That would also deserve a note in the docs that `use_only_cwd_commits` will be ignored unless `version_source = "commit"` too.

Ideally we'd implement a full solution for #168 but that's a lot more effort - before that happens, I think this is an ok halfway house.

kibidev:
@bernardcooke53 thanks for the suggestions, I have updated the code and added a bit more documentation now.

kibidev:
I have (finally) revisited this PR and evaluated the possibility to add tests.
It seems to me that the relevant function (get_commit_log from vcs_helpers.py) is currently tested only on this repository.
To test the functionality from this PR in this way is not possible, as this is not a mono repo (or have suitable sub folders to get predictable commit logs from).
In most other relevant tests, get_commit_log itself is mocked, and so it would not make sense to test this PR in a similar way.

The remaining approach would be to write new tests that would mock git.Repo or similar to test get_commit_log. In my opinion the value from this would also be minimal, as it would mostly just test the mock.

I will probably not add any testing to this PR, for the way forward I can suggest one of:
- Close this PR and do not add the functionality (I'll use a fork for the functions I need)
- Merge without adding more tests
- Someone suggest a concrete way to tests the functionality, and I'll write the tests

@bernardcooke53

bernardcooke53:
Thanks for coming back to this @kibidev - I agree with your assessment, it's a pain to test this thoroughly with the current tests.
Given it's a relatively simple change I think it's ok to merge - but I'm not going to include this in v8, at least initially, I really want monorepo support to be done "properly"

PR #511, opened 2022-10-21
ci: Update deprecated actions

kibidev:
Update actions to remove some warnings about deprecated features.


danth (approved):
Looks good to me.

Perhaps we could configure Dependabot to update the actions automatically?

PR #518, opened 2022-11-01
feat: Update action with configuration options

cmitzel-ncino:
Updating the action to use configured git environment variables according to [vanilla semantic release](https://semantic-release.gitbook.io/semantic-release/usage/configuration) and [official git documentation](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing).

### DIFF
--- a/action.sh
+++ b/action.sh
@@ -13,7 +13,7 @@ export PATH="${PATH}:/semantic-release/.venv/bin"
 cd "${INPUT_DIRECTORY}"
 
 export GIT_COMMITER_NAME="${INPUT_GIT_COMMITER_NAME:="github-actions"}"
-export GIT_COMMITER_EMAIL="${GIT_COMMITER_EMAIL:="github-actions@github.com"}"
+export GIT_COMMITER_EMAIL="${INPUT_GIT_COMMITER_EMAIL:="github-actions@github.com"}"
 
 # Set Git details
 git config --global user.name "$GIT_COMMITER_NAME"
--- a/action.sh
+++ b/action.sh
@@ -8,12 +8,11 @@ export PYPI_TOKEN="${INPUT_PYPI_TOKEN}"
 export REPOSITORY_USERNAME="${INPUT_REPOSITORY_USERNAME}"
 export REPOSITORY_PASSWORD="${INPUT_REPOSITORY_PASSWORD}"
 export PATH="${PATH}:/semantic-release/.venv/bin"
+export GIT_COMMITER_NAME="${INPUT_GIT_COMMITER_NAME:="github-actions"}"
+export GIT_COMMITER_EMAIL="${INPUT_GIT_COMMITER_EMAIL:="github-actions@github.com"}"
 
 # Change to configured directory
 cd "${INPUT_DIRECTORY}"
-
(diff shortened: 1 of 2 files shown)


action.sh
```diff
@@ -12,13 +12,16 @@ export PATH="${PATH}:/semantic-release/.venv/bin"
 # Change to configured directory
 cd "${INPUT_DIRECTORY}"
 
+export GIT_COMMITER_NAME: "${GIT_COMMITER_NAME:="github-actions"}"
+export GIT_COMMITER_EMAIL: "${GIT_COMMITER_EMAIL:="github-actions@github.com"}"
```
Kevmo92, line 16:
```suggestion
export GIT_COMMITER_NAME="${GIT_COMMITER_NAME:="github-actions"}"
export GIT_COMMITER_EMAIL="${GIT_COMMITER_EMAIL:="github-actions@github.com"}"
```

Kevmo92 (commented):
export command should be updated

Kevmo92 (approved):
Nice

https://media.giphy.com/media/yJFeycRK2DB4c/giphy.gif

relekang (changes requested):
According to the [github actions documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations) this won't work:
>Any environment variables set in an env context defined at the workflow level in the caller workflow are not propagated to the called workflow.

This needs to be configured to with [action inputs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#inputs)

relekang:
And btw the documentation you are referring to is the node semantic release and not python semantic release. Our documentation is on https://python-semantic-release.readthedocs.io/en/latest/

cmitzel-ncino:
@relekang does that look right now?

And yeah, I was under the impression that the projects were related in some way but I must have misunderstood. 

We're trying to just make sure that we can create a commit and release signed by our service user rather than using the general actions user.

action.sh
```diff
@@ -12,13 +12,16 @@ export PATH="${PATH}:/semantic-release/.venv/bin"
 # Change to configured directory
 cd "${INPUT_DIRECTORY}"
 
+export GIT_COMMITER_NAME="${GIT_COMMITER_NAME:="github-actions"}"
```
relekang, line 15:
```suggestion
export GIT_COMMITER_NAME="${INPUT_GIT_COMMITER_NAME:="github-actions"}"
```

action.sh
```diff
@@ -12,13 +12,16 @@ export PATH="${PATH}:/semantic-release/.venv/bin"
...
+export GIT_COMMITER_EMAIL="${GIT_COMMITER_EMAIL:="github-actions@github.com"}"
```
relekang, line 16:
```suggestion
export GIT_COMMITER_EMAIL="${INPUT_GIT_COMMITER_EMAIL:="github-actions@github.com"}"
```

action.sh
cmitzel-ncino: good catch, i merged in kevmo's changes wrong :oof:

relekang (approved):

Kevmo92 (approved):
Thanks!

relekang:
Thanks @cmitzel-ncino 🙌

PR #520, opened 2022-11-01
fix(action): fix environment variable names

cmitzel-ncino:
My previous contribution had some incorrect variable names that I didn't catch on first glance because english even as a native speaker is a silly silly language.


cmitzel-ncino:
@relekang I've fixed the names of the environment variables to match that of the variables in the [git documentation],(https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing) apologies for not double checking.

relekang (approved):

PR #522, opened 2022-11-02
feat(repository): fixes #521, adding support for TWINE_CERT

### LINKED ISSUE #521: Add Support for TWINE_CERT
## Description

TWINE_CERT is an environment variable which allows using self-signed certificate bundles for SSL verification to repositories. Would be helpful to pass this value along in the twine settings if it's set.

## Use cases

Self signed cert bundles are commonly used in corporate settings. 

## Possible implementation

Check for existence of TWINE_CERT, verify it points to a valid file, and include in the twine settings if so.

tyschroed:
Adds support for TWINE_CERT, as proposed [here](https://github.com/python-semantic-release/python-semantic-release/issues/521).

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -159,3 +159,7 @@ Custom repository (package index) URL to upload the package to.
 Takes precedence over :ref:`config-repository_url`
 
 See :ref:`automatic-dist-upload` for more about uploads to custom repositories.
+
+``TWINE_CERT``
+------------------
+Path to alternative CA bundle to use for SSL verification to repository. `See here for more information <https://twine.readthedocs.io/en/stable/>`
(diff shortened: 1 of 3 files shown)


bernardcooke53 (commented):
LGTM but would be nice to have an explicit mention in [the docs](https://github.com/python-semantic-release/python-semantic-release/blob/master/docs/index.rst#distributing-release-on-pypi-or-custom-repository)
I have a sense of foreshadowing that someone else may need this and without changing the docs would not know how to do it :slightly_smiling_face:

tyschroed:
Good call - updated the docs!

bernardcooke53 (approved):
LGTM👍

PR #523, opened 2022-11-04
feat: add signing options to action

cmitzel-ncino:
Add ability to add an SSH key for verifying commits as mentioned here: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification

Updated documentation accordingly.

Install `stable-bpo` version of `git` ([1:2.34.1-1~bpo11+1](https://packages.debian.org/source/stable-backports/git)) in docker image as the stable version ([1:2.30.2-1](https://packages.debian.org/source/stable/git)) doesn't support signing commits with SSH. 

Added documentation for the action now that there are more than a few options.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,11 @@ RUN set -ex; \
     apt-get install -y --no-install-recommends \
         git-lfs
 
+#install backported stable vesion of git, which supports ssh signing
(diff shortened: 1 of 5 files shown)


cmitzel-ncino:
@relekang following up on the last contribution I made, our security team wants us to verify commits from our service accounts. Hoping I managed to implement it in a way that's general enough that others can make use of it as well.

relekang (approved):

PR #525, opened 2022-11-14
fix: unset mypy version

mrjk:
When I forked this repo and run my own github action, the test failed because numpy reported many errors. While looking back at this original project, the CI tests build are working well with the version of `mypy-0.982` but on my side but I got the latest on my side, which reports more errors and make tests to fail.

As the test has been written for a version of `mypy-0.982`, I propose to fix the version in the attendances.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
         ],
         "docs": ["Sphinx==1.3.6", "Jinja2==3.0.3"],
         "dev": ["tox", "isort", "black"],
-        "mypy": ["mypy", "types-requests"],
+        "mypy": ["mypy==0.982", "types-requests"],
     },
     entry_points="""
         [console_scripts]


relekang:
Was these errors fixed by #527?

bernardcooke53:
Hey @mrjk , I agree with you that fixing dependencies is generally good practice, though because this isn't something in the runtime it's not as important (e.g. you can just avoid a buggy version or pin below a breaking change when one's released, noone who installs PSR will be affected by that). Fixing the type-hinting is the way to go, though - it would be good to see exactly which mypy errors you got on 0.982 compared to 0.991.

I've been using 0.990/0.991 doing some refactoring work on PSR without any issues. The reason #527 fixed this is because in mypy 0.980 the configuration setting `implicit_optional` was `True` by default, and now it's `False` - so it would have been no loss to just change the config for mypy, though as you said it's preferable to add the correct type hints as you did in your PR :slightly_smiling_face:

I don't think for the development/testing dependencies it's worth the overhead of keeping versions up to date by hand - maybe adding dependabot would be a good step in the future?

mrjk:
From my contributor POV, I just like when things work out of the box, so you take that risk to discourage some potential contributor, when things break down. On the other side, it forces you to track on mypy good coding practices, which is also good, until the project is alive.

I'm fine with both options, until the project is alive, no need to fix deps; if it stale at some point, it will be good to fix the version for later use.

I let you choose what you want to do with this PR ;)

bernardcooke53:
I think we can let this stay unpinned for the time being, if I manage to get to it I'll try and set dependabot up and then we can both pin _and_ get some automated dependency bumping

PR #527, opened 2022-11-14
fix: fix mypy errors for publish

cmitzel-ncino:
The publish process is held up by errors in the mypy step so I've fixed the errors to the best of my ability. 

Screen Shot 2022-11-14 at 11 49 34 AM


cmitzel-ncino:
@relekang this should fix the issues with the publish process :D

relekang:
Thanks 🙌

PR #546, opened 2023-01-17
fix: if prerelease, publish prerelease (python-semantic-release#544)

wintero92:
* Fixes publushing process of release/prerelease

Running `semantic-release publish --prerelease` currently creates release.


ak4wa1:
What happened? Did you find a solution @wintero92 ? I am going through the same situation

PR #582, opened 2023-04-24
Update Gitmojis according to official node module and make utf-8 symbols available

sr-verde:
According to the "official" [node module](https://github.com/carloscuesta/gitmoji/blob/master/packages/gitmojis/src/gitmojis.json) the Gitmojis have changed. This PR updates the default emojis for patch and minor versions.

In addition, it allows the usage of utf-8 symbols. So one can either use `:sparkles` or `✨` in their commit messages.


danth (approved):
Thanks for updating this, there have been several new emojis since Gitmoji support was added.

PR #585, opened 2023-04-29
Doesn't stage CHANGELOG.md if version_source configuration option is tag_only

### LINKED ISSUE #381: Add option to skip writing changelog to file
## Description

Add an option to skip writing changelog file to disk or skip adding it to the VCS at least.

## Use cases

I would like to use semantic-release together with setuptools_scm. I want to use semantic release to bump the version tag as a "push to main" action in CI but not make any file changes. The version is extracted from the git tag by setuptools_scm, but if the repo state is dirty, it will be counted as a dev-version.

If CHANGELOG.md is added to gitignore to make sure it doesn't count towards the repo dirty state, the `git add` operation will raise an error because semantic-release is trying to add an ignored file.

## Possible implementation

Add an option to explicitly skip writing the changelog to file (or skip adding it to git)

sim6:
Changelog and additional files should not be staged by publish() with update_changelog_file() and
update_additional_files() if the `version_source` configuration option is `tag_only` due to no
commit is added by bump_version().

Fixes #381

### DIFF
--- a/semantic_release/cli.py
+++ b/semantic_release/cli.py
@@ -369,8 +369,9 @@ def publish(
             run_pre_commit()
 
         if not retry:
-            update_changelog_file(new_version, changelog_md)
-            update_additional_files()
+            if config.get("version_source") != "tag_only":
+                update_changelog_file(new_version, changelog_md)
+                update_additional_files()
             bump_version(new_version, level_bump)
         # A new version was released
         logger.info("Pushing new version")
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -928,6 +928,59 @@ def test_publish_should_call_functions(mocker):
     mock_checkout.assert_called_once_with("master")
 
 
+def test_publish_by_tag_only_should_not_call_update_functions(mocker):
+    mocker.patch(
+        "semantic_release.cli.config.get",
+        wrapped_config_get(version_source="tag_only"),
+    )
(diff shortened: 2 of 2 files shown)


sim6:
Commit message typo fixed.

sim6:
Commit messages improved.

sim6:
Commit messages improved again :(

bernardcooke53:
Just got round to taking a look at this @sim6 - thanks for putting a PR together! :rocket:

However I'm not sure this is the way to fix the issue - I checked the [docs](https://python-semantic-release.readthedocs.io/en/latest/configuration.html#version-source) and maybe the wording is _ambiguous_, but it does state:

> If set to tag_only, then version_variable is ignored and no changes are made or committed to local config files. The current version from the latest tag matching vX.Y.Z. This won’t change the source defined in [version_variable](https://python-semantic-release.readthedocs.io/en/latest/configuration.html#config-version-variable).

I wouldn't consider a changelog a config file... I might be irritated if I had a colleague that used a changelog as a config file, too :grin: what I'm getting at is - I think it's currently expected behaviour that PSR will manage the changelog file regardless of `version_source`, so this would be a breaking change. I would prefer this done with an explicit config setting like `enable_changelog` or something along those lines (with default `true` to get the same behaviour as now).

I'm going to close this - not because the effort isn't appreciated, just because I want to avoid those issues above. I think this code is a start towards implementing the new option though so if you want to raise a new PR later, please do. Thanks again!

PR #547, opened 2023-01-17
fix: if prerelease, publish prerelease (python-semantic-release#544)

wintero92:
* Fixes publishing process of release/prerelease

Running `semantic-release publish --prerelease` currently creates release, not prerelease.

PR #587, opened 2023-05-03
fix: if prerelease, publish prerelease (python-semantic-release#544) (continuing #547)

tofuurem:
Since @wintero92 will not be able to complete his PR #547 in the next couple of months, I have corrected the tests so that we can already see this version in release.

### DIFF
--- a/tests/test_hvcs.py
+++ b/tests/test_hvcs.py
@@ -408,7 +408,7 @@ class GithubReleaseTests(TestCase):
 
             with mock.patch.dict(os.environ, {"NETRC": netrc_file.name}):
                 status = Github.post_release_changelog(
-                    "relekang", "rmoq", "1.0.0", "text"
+                    "relekang", "rmoq", "1.0.0", "text", False
                 )
                 self.assertTrue(status)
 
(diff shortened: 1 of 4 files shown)


tests/test_hvcs.py
```diff
@@ -202,11 +201,60 @@ def test_get_domain_should_have_expected_domain(
...
-
             assert get_hvcs().domain() == expected_domain
             assert get_hvcs().api_url() == expected_api_url
 
 
+@responses.activate
+@pytest.mark.parametrize(
+    "hvcs,url,prerelease",
+    [
+        ("github", "https://api.github.com/repos/relekang/rmoq/releases", True),
```
bernardcooke53, line 212: I think it'd be cleaner passing the class in as `hvcs` instead of a string, and if you do two `parametrize` decorators you can get the same effect with less params per test; i.e.

```python
@pytest.mark.parametrize(
    "hvcs, url",
    [
        (Github, "https://api.github.com/repos/relekang/rmoq/releases"),
        (Gitea, "https://gitea.com/api/v1/repos/gitea/tea/releases")
    ]
)
@pytest.mark.parametrize("prerelease", (True, False))
def test_should_create_prerelease_if_asked_for(hvcs, url, prerelease):
    ...
```

That (+ the netrc line) might have cut down on the `if`s in the test itself - but I don't see anything wrong with what you've done here :+1:

bernardcooke53 (approved):
Thanks @wintero92 and @tofuurem :tada:

wintero92:
Thanks for stepping in @tofuurem

PR #590, opened 2023-05-18
fix: update docs and default config for gitmoji changes

cowofevil:
PR #582 updated to the latest Gitmojis release however the documentation and default config options still referenced old and unsupported Gitmojis.


danth (approved):
Thanks!

PR #594, opened 2023-05-26
fix: generate markdown linter compliant changelog headers & lists

ringohoffman:
Adding an extra new line after each header and between sections fixes the following 2 `markdownlint` ([VS Code](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint), [GitHub](https://github.com/DavidAnson/markdownlint)) errors for changelogs generated by this package:  
* [`MD022` - Headings should be surrounded by blank lines](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022---headings-should-be-surrounded-by-blank-lines)
* [`MD032` - Lists should be surrounded by blank lines](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md032---lists-should-be-surrounded-by-blank-lines)

PR #597, opened 2023-05-30
fix: generate markdown linter compliant changelog headers & lists

ringohoffman:
In #594, I missed that there are 2 places where the version header is formatted, one with dates and one without.

In this PR, I add the extra newline to make the generated changelogs `markdownlint` compliant when the date is included in the header.


ringohoffman:
@bernardcooke53 does this look okay?

bernardcooke53 (approved):
Yep - looks good to me

Issue #602, opened 2023-06-12
8.0.0a7 Inconsistent types of assets with pydantic


smeng9:
## The problem

The types for assets are not consistent and users cannot upload assets to vcs

## Expected behavior

The types for assets in config should be consistent

## Additional context

In [configurations.rst](https://github.com/python-semantic-release/python-semantic-release/blob/8.0.x/docs/configuration.rst#assets-liststr) assets is List[str]

In [config.py](https://github.com/python-semantic-release/python-semantic-release/blob/8.0.x/semantic_release/cli/config.py#L129) assets is Tuple[str, ...]

In [version.py](https://github.com/python-semantic-release/python-semantic-release/blob/8.0.x/semantic_release/cli/commands/version.py#L319) It is trying to concat tuple to a list which causes an exception.

bernardcooke53:
Closed by #603

PR #603, opened 2023-06-12
fix: correct assets type in configuration

smeng9:
Closes https://github.com/python-semantic-release/python-semantic-release/issues/602

Also do we have an ETA of 8.0 release?


kedvall:
Same issue here, any eta for this fix being released?

PR #612, opened 2023-06-26
feat: Ignore commits by message prefix

sr-verde:
I need to ignore some commits by a prefix of their message.

This MR adds a check whether to add or to ignore a commit by the prefix of its message. I added that in `get_commit_log` to ignore them as soon as possible.

This MR still needs unit testing, but I am happy to hear your opinion.


sr-verde:
Feedback welcome.

PR #617, opened 2023-07-05
feat(commit): add option to filter commits for monorepos

roggervalf:
I'm working on a monorepo, where I have shared files and I would like to differentiate which commits affect my components having an option to filter my commits, so only those commits should appear in the release notes. ref https://github.com/python-semantic-release/python-semantic-release/issues/168


roggervalf:
Please @bernardcooke53 when you get some time

PR #618, opened 2023-07-05
feat(deps): bump semver to version 3

roggervalf:
bumping semver to version 3, it looks like methods that are used in this project are compatible


bernardcooke53:
Thank you for the PR @roggervalf ! After the last major release, we no longer depend on semver, so I'm going to close this - but the effort is very much appreciated :rocket:

PR #643, opened 2023-07-18
docs: clarify v8 breaking changes in GitHub action inputs

### LINKED ISSUE #642: Clarifications on GitHub actions inputs changes in v8
The "Migrating from Python Semantic Release v7" page has a section describing some breaking changes around the inputs, but I don't fully understand it. Here is where I block:

1. The removal of the PyPI token is clear, but [the alternative](https://github.com/browniebroke/python-semantic-release/blob/2129b729837eccc41a33dbb49785a8a30ce6b187/docs/migrating_from_v7.rst?plain=1#L118-L121) mentions using `repository_username`:

   > You should supply the token as the value of the ``repository_username`` instead, and if using only the token for authentication, remember to set the ``repository_username`` input equal to ``__token__``.

   This was [visibly removed](https://github.com/python-semantic-release/python-semantic-release/pull/619/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6). I'm not sure what the author meant to write here?
   
2. The [following section](https://github.com/browniebroke/python-semantic-release/blob/2129b729837eccc41a33dbb49785a8a30ce6b187/docs/migrating_from_v7.rst?plain=1#L125C3-L125C21) mentions a change of purpose of `additional_options`, but it was not only repuposed, it was also [renamed](https://github.com/python-semantic-release/python-semantic-release/pull/619/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6).

3. the same section mentions 2 inputs, `version_options` and `publish_options`, which aren't actually present in [the `action.yml` file](https://github.com/browniebroke/python-semantic-release/blob/master/action.yml).

I'll try to send a PR to fix these. Let me know if I'm missing something.

browniebroke:
Fix #642

### DIFF
--- a/docs/migrating_from_v7.rst
+++ b/docs/migrating_from_v7.rst
@@ -113,24 +113,24 @@ GitHub Action:
 
 .. _breaking-github-action-removed-pypi-token:
 
-Removal of ``pypi_token`` input
-"""""""""""""""""""""""""""""""
-The ``pypi_token`` input of the GitHub action has been removed. You should
-supply the token as the value of the ``repository_username`` instead,
-and if using only the token for authentication, remember to set the
-``repository_username`` input equal to ``__token__``.
+Removal of ``pypi_token``, ``repository_username`` and ``repository_password`` inputs
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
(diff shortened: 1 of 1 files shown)


bernardcooke53 (approved):
Thank you - agree this is a clearer explanation

PR #644, opened 2023-07-18
fix: handle missing configuration

bernardcooke53:
With this change, PSR will no longer error if it cannot find appropriate configuration in the default
pyproject.toml file, or in the file specified by -c/ --config when that option is given. Instead it will use the default configuration values

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 readme = "README.rst"
 authors = [{ name = "Rolf Erik Lekang", email = "me@rolflekang.com" }]
 dependencies = [
-  "click>=7,<9",
+  "click>=8,<9",
   "gitpython>=3.0.8,<4",
   "requests>=2.25,<3",
   "jinja2>=3.1.2,<4",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ classifiers = [
 readme = "README.rst"
 authors = [{ name = "Rolf Erik Lekang", email = "me@rolflekang.com" }]
 dependencies = [
-  "click>=7,<9",
+  "click>=8,<9",
```
bernardcooke53, line 24: don't allow previous major versions of `click`

pyproject.toml
```diff
@@ -62,7 +62,7 @@ test = [
   "types-pytest-lazy-fixture>=0.6.3.3",
 ]
 dev = ["tox", "isort", "black"]
-mypy = ["mypy", "types-requests", "types-click"]
+mypy = ["mypy", "types-requests"]
```
bernardcooke53, line 65: `types-click` was actually breaking mypy for me locally - I suppose it's no longer updated with click v8 (hence pinning above)

pyproject.toml
```diff
@@ -124,7 +124,7 @@ commands =
 [testenv:mypy]
 deps = .[mypy]
 commands =
-    mypy --ignore-missing-imports semantic_release
+    mypy semantic_release
```
bernardcooke53, line 127: no need for this flag any more

Issue #639, opened 2023-07-18
`semantic-release version --prerelease` does not force a prerelease version bump.


css-optivoy:
## The problem

`semantic-release version --prerelease` does not force a prerelease version bump:

```
$ git log
commit 2f01c54100ae5a63020e267a397221f088f927e3 (HEAD -> bump)
Author: css
Date:   Tue Jul 18 10:04:41 2023 +0200

    build(meh): Blah

commit bfed5b9fda67e7ef3e251c18860f48b59798b7a8 (tag: 0.27.7-dev.3)
Author: semantic-release <semantic-release>
Date:   Mon Jul 17 11:33:02 2023 +0000

    0.27.7-dev.3
    
    Automatically generated by python-semantic-release

$ semantic-release version --prerelease
[...]
           WARNING  [semantic_release.cli.commands.version] WARNING version.version: Forcing prerelease due to '--prerelease' version.py:247
                    command-line flag     
           INFO     [semantic_release.version.algorithm] INFO algorithm.tags_and_versions: found 303 previous tags           algorithm.py:55
           INFO     [semantic_release.version.algorithm] INFO algorithm.next_version: Found 106 full releases (excluding    algorithm.py:245
                    prereleases)                                                                                                            
           INFO     [semantic_release.version.algorithm] INFO algorithm.next_version: The last full release was 0.27.6,     algorithm.py:267
                    tagged as '0.27.6'                                                                                                      
           INFO     [semantic_release.version.algorithm] INFO algorithm.bfs: found latest version in branch history:         algorithm.py:91
                    '0.27.6' (f5dbb72)                                                                                                      
           INFO     [semantic_release.version.algorithm] INFO algorithm._bfs_for_latest_version_in_history: the latest      algorithm.py:110
                    version in this branch's history is 0.27.6                                                                              
           INFO     [semantic_release.version.algorithm] INFO algorithm.next_version: The last full version in this         algorithm.py:291
                    branch's history was 0.27.6                                                                                             
           INFO     [semantic_release.version.algorithm] INFO algorithm.next_version: The type of the next release release  algorithm.py:368
                    is: no_release                                                                                                          
           INFO     [semantic_release.version.algorithm] INFO algorithm.next_version: No release will be made               algorithm.py:370
0.27.7-dev.3
           INFO     [semantic_release.version.algorithm] INFO algorithm.tags_and_versions: found 303 previous tags           algorithm.py:55
[...]

PR #647, opened 2023-07-19
feat(cli): correct --prerelease use and add --as-prerelease

bernardcooke53:
Prior to this change, --prerelease performed the role that --as-prerelease now does, which was an unintentional breaking change to the CLI compared to v7.

--prerelease now forces the next version to increment the prerelease revision, which makes it consistent with --patch, --minor and --major, while --as-prerelease allows for the usual next version to be converted to a prerelease befoer it is applied to the project

Resolves: #639

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -173,38 +173,109 @@ number (``1.0.0``).
 
 .. _cmd-version-option-force-level:
 
-``--major/--minor/--patch``
-***************************
+``--major/--minor/--patch/--prerelease``
+****************************************
 
-Force the next version to increment the major, minor or patch digit, respectively.
-These flags are optional but mutually exclusive, so only one may be supplied, or none at all.
-Using these flags overrides the usual calculation for the next version; this can be useful, say,
-when a project wants to release its initial 1.0.0 version.
+Force the next version to increment the major, minor or patch digits, or the prerelease revision,
+respectively. These flags are optional but mutually exclusive, so only one may be supplied, or
+none at all. Using these flags overrides the usual calculation for the next version; this can
+be useful, say, when a project wants to release its initial 1.0.0 version.
 
 .. warning::
-    Using these flags will override the value of prerelease, **regardless of your configuration or the
-    current version**. To produce a prerelease with the appropriate digit incremented you should also
-    supply the :ref:`cmd-version-option-prerelease` flag. If you do not, using these flags will force
+
+    Using these flags will override the configured value of ``prerelease`` (configured
+    in your :ref:`Release Group<multibranch-releases-configuring>`),
+    **regardless of your configuration or the current version**.
+
+    To produce a prerelease with the appropriate digit incremented you should also
+    supply the :ref:`cmd-version-option-as-prerelease` flag. If you do not, using these flags will force
     a full (non-prerelease) version to be created.
+
+For example, suppose your project's current version is ``0.2.1-rc.1``. The following
+shows how these options can be combined with ``--as-prerelease`` to force different
+versions:
+
+.. code-block:: bash
+
+   semantic-release version --prerelease --print
+   # 0.2.1-rc.2
+
+   semantic-release version --patch --print
+   # 0.2.2
+
+   semantic-release version --minor --print
+   # 0.3.0
+
+   semantic-release version --major --print
+   # 1.0.0
+
+   semantic-release version --minor --as-prerelease --print
+   # 0.3.0-rc.1
+
+   semantic-release version --prerelease --as-prerelease --print
+   # 0.2.1-rc.2
 
 These options are forceful overrides, but there is no action required for subsequent releases
 performed using the usual calculation algorithm.
+
+Supplying ``--prerelease`` will cause Python Semantic Release to scan your project history
+for any previous prereleases with the same major, minor and patch versions as the latest
+version and the same :ref:`prerelease token<cmd-version-option-prerelease-token>` as the
+one passed by command-line or configuration. If one is not found, ``--prerelease`` will
+produce the next version according to the following format:
+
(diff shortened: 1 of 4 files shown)


css-optivoy:
I'm no expert, but something feel a little off about this logic to me.

This example from the PR seems to violate SemVer:
> if `0.2.1-rc.1` already exists as a previous version, and the latest version is `0.2.1`, invoking `semantic-release version --prerelease --prerelease-token "rc"` will produce `0.2.1-rc.2`

Albeit the section that follows it warns about that.

I haven't tested this, but I feel like PSR v7, when given `--prerelease`, would first apply its automatic version bump logic to determine the next version, and then apply the prerelease tag.
I.e. `0.2.1-rc.1` would become `0.2.2-rc.1` if only patch changes had happened since `0.2.1`.


The other thing I'm having a hard time wrapping my head around is the `--prerelease`/`--as-prerelease` duality. I had to read through the PR a couple of times to understand what they each do, and why we might need them.

My understanding is that `--as-prerelease` is there primarily to allow us to go from e.g. `0.1.1` directly to `0.2.0-rc.1` via `--minor --as-prerelease`. And this is in turn because we can't use `--minor --prerelease` for that since all the bump flags are mutually exclusive.


Now, the behaviour I would expect is something closer to this:
1. `--prerelease` forces the next prerelease version. Either by bumping the prerelease version if the latest version is a prerelease (`1.2.3-pre.1` -> `1.2.3-pre.2`), or by bumping the version, and then tagging that as a prerelease (`1.2.3` -> `1.2.4-pre.1`).
2. The other bump flags, `--major`, `--minor` and `--patch`, can be combined with `--prerelease` to force a manual bump to a greater prerelease, so e.g. `--major --prerelease` would give you `1.8.9` -> `2.0.0-pre.1`.

This should largely preserve existing behaviour, and avoid the above potential issues.

How well this works with the new multi-branch versioning I can't really say, as I unfortunately have no experience with such a setup.

bernardcooke53:
Thanks for taking a look @css-optivoy. It's kind of gnarly logic, I admit. Let me try to respond thoroughly to each of your comments:

> This example from the PR seems to violate SemVer:
> 
> if 0.2.1-rc.1 already exists as a previous version, and the latest version is 0.2.1, invoking semantic-release version --prerelease --prerelease-token "rc" will produce 0.2.1-rc.2
> 
> Albeit the section that follows it warns about that.

I don't think it violates SemVer, it's just that it's possible to shoot yourself in the foot if you're developing on a branch that's out-of-date compared to your main/latest release. SemVer doesn't mandate that your project versions are released monotonically, only that according to the precedence rules certain versions are "higher" than others. I can release v1.0.1, v2.0.0 and then v1.1.0 in that order, if need be - it's just everyone looking for "latest" by semver will get v2.0.0 instead of v1.1.0. 

Let me try to clarify the example (and then update the docs):

If you release normal versions from `main` and prereleases from `feat/*` branches and you last released `0.2.0`, suppose you do the following:
* create `feat/A` and `feat/B` off of `main`
* develop a `fix` on `feat/A`,  do a prerelease to get `0.2.1-rc.1`
* merge `feat/A` into `main`, and do a release to get `0.2.1`
* commit a `fix` to `feat/B` and try to do a prerelease

Now, from this prerelease, the commit history from the point of view of HEAD on `feat/B` only knows of `0.2.0` as the last major version - so a patch increment + prerelease would result in `0.2.1-rc.1`. Obviously that tag already exists so there's two choices:
* either error out/don't release, indicating that `0.2.1-rc.1` is already released, or
* recognise that `0.2.1-rc.1` is already released and increment to `0.2.1-rc.2`

Having to deal with multiple branches makes this a pain to get right, however it's not possible to naively search the repository tags and pick the latest version from those to base the next version off of - if you pick up a branch from 3 years ago, you'll be prereleasing for the incremented version with a wildly stale codebase. It would prevent trying to support multiple versions.

I was/am still considering whether the tag-search for prereleases leading up to the same normal version is the "correct" thing to do. It would be much easier to take the stance of "use build metadata for this instead of prerelease tags", which is honestly a more robust way to go about it, but that would in turn ask for a change in versioning scheme from (I imagine) a lot of users.

___
> I haven't tested this, but I feel like PSR v7, when given --prerelease, would first apply its automatic version bump logic to determine the next version, and then apply the prerelease tag.
I.e. 0.2.1-rc.1 would become 0.2.2-rc.1 if only patch changes had happened since 0.2.1.

This is technically true, but it was influenced by the default for `--prerelease-patch` rather than `--prerelease` directly. In my opinion, if you're using flags on the command-line to alter the behaviour of the next version from what the commits would otherwise indicate, the most important thing is having the level of control to decide what your next version is.

For example, if I want to run an ad-hoc prerelease from `main`, I can use `--as-prerelease` to calculate my next version from the commits that need releasing, and then ensure I publish a prerelease version. If I want to force a minor prerelease, I can use `--minor --as-prerelease` to reach that version. Through combination of `--prerelease/--patch/--minor/--major` I can force the increment of any of the version digits that I need to, and through `--as-prerelease` I can ensure I construct a prerelease either _if_ I need to release a new version (with no other forcing flags) or intentionally along with a forced increment.

What could potentially be missing here is a `--minimum-bump="none|patch|minor|major"` flag, with "bump by at least x level" semantics.

I think the usage you suggested with `--prerelease` and `--patch --prerelease` has a similar risk of confusion as `--prerelease` vs `--as-prerelease` imo - the behaviour of the flag would change depending on the _other_ flags provided, and your point 1 imposes an explicit patch bump that can't be disabled (something maybe a `--minimum-bump` flag could mitigate). `--as-prerelease` is new functionality and I intended it to be there for more of that granular control over the next version.

Definitely open to renaming it if you have a clearer/more explicit name suggestion!

PR #651, opened 2023-07-21
docs: add Python 3.11 to classifiers in metadata

afuetterer:
Hi there, my first PR to the project.

The tests in GitHub Actions run on Python 3.11 as well. I guess you should add it to the project metadata then.


bernardcooke53:
Thanks @afuetterer!

PR #653, opened 2023-07-22
fix(changelog): use base version without tag format by default

roggervalf:
when release notes are generated, tag version is set instead of baser version. Before breaking change we were having base version format "v{version}" that reflects the actual published version. I think tag formatting is used for avoiding multi packages conflicts in the same repo.

### DIFF
--- a/semantic_release/version/version.py
+++ b/semantic_release/version/version.py
@@ -205,6 +205,9 @@ class Version:
     def as_tag(self) -> str:
         return self.tag_format.format(version=str(self))
 
+    def as_base_version(self) -> str:
+        return "v{version}".format(version=str(self))
+
     def bump(self, level: LevelBump) -> Version:
         """
         Return a new Version instance according to the level specified to bump.
--- a/semantic_release/version/version.py
+++ b/semantic_release/version/version.py
@@ -205,7 +205,7 @@ class Version:
     def as_tag(self) -> str:
         return self.tag_format.format(version=str(self))
 
-    def as_base_version(self) -> str:
+    def as_semver_tag(self) -> str:
         return "v{version}".format(version=str(self))
 
     def bump(self, level: LevelBump) -> Version:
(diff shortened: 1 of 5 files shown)


roggervalf:
please @bernardcooke53 when you get some chance :eyes:

bernardcooke53:
Was just coming back to this @roggervalf - I'm not the biggest fan of having "one special format" but I agree it's better to keep the old behaviour in place. Functionally the code looks good - thank you for the PR :pray:

I don't think "base_version" is a good name for the method though, I think we should call it `as_semver_tag` as the convention derives from [SemVer 1.0.0](https://semver.org/spec/v1.0.0.html#tagging-specification-semvertag). It will be usable by template authors so better to be explicit with naming.

roggervalf:
sounds good to me, let me update it

semantic_release/version/version.py
```diff
@@ -205,6 +205,9 @@ def __repr__(self) -> str:
     def as_tag(self) -> str:
         return self.tag_format.format(version=str(self))
 
+    def as_base_version(self) -> str:
+        return "v{version}".format(version=str(self))
```
bernardcooke53, line 209: (just an fyi) you can actually just `return f"v{self}", f-strings will automatically format `self` with `self.__str__()` if it's defined, and if not it will try to format with `self.__repr__()` - from the [docs](https://docs.python.org/3/reference/datamodel.html#object.__str__):

> Called by [str(object)](https://docs.python.org/3/library/stdtypes.html#str) and the built-in functions [format()](https://docs.python.org/3/library/functions.html#format) and [print()](https://docs.python.org/3/library/functions.html#print) to compute the “informal” or nicely printable [string](https://docs.python.org/3/library/stdtypes.html#textseq) representation of an object. The return value must be a string object.

and from the [PEP](https://peps.python.org/pep-0498) , f-string call the builtin `format` method on expressions:

> The expression is then formatted using the __format__ protocol, using the format specifier as an argument. The resulting value is used when building the value of the f-string.

> Note that __format__() is not called directly on each value. The actual code uses the equivalent of type(value).__format__(value, format_spec), or format(value, format_spec). See the documentation of the builtin format() function for more details.

roggervalf:
@bernardcooke53 it should be addressed

bernardcooke53 (approved):
LGTM

Issue #652, opened 2023-07-21
Changelog placeholder is ignored


bernardcooke53:
## The problem

It looks like the placeholder  is ignored, the changelog is regenerated with all the previous commits, and not only taking the last ones, as an example https://github.com/taskforcesh/bullmq/commit/59d4d1dd67a0efbab6ff252b2a042fc61c3f4df2

## Environment

Please state which OS you are using and provide the output of the following commands:

```shell
python --version
pip --version
semantic-release --version
pip freeze
```

Python 3.10.12
pip 23.2
PSR 8.0.3

Please also indicate which Python build tool(s) you are using (e.g. `pip`, `build`,
`poetry`, etc.), including the version number too.

pipenv 2023.7.4 + setuptools 68

🛠 `Running build command: python3 setup.py sdist`

## Configuration

https://github.com/taskforcesh/bullmq/tree/master/python

## Logs

Action: https://github.com/taskforcesh/bullmq/actions/runs/5624948662/job/15242766976

codejedi365:
Duplicate with #858 & #722

codejedi365:
I have found a solution that works for re-introducing the placeholder version update for changelogs using jinja. It will be released as a custom template configuration first in `v9` and then converted back to the default in `v10`. Stay tuned.

PR #664, opened 2023-07-26
fix(changelog): consider placeholder for old version

roggervalf:
ref https://github.com/python-semantic-release/python-semantic-release/issues/652


roggervalf:
hi @bernardcooke53 , do you know if there are test cases to cover the case where template_dir does not exist?

bernardcooke53:
Hi @roggervalf - apologies, I've been quite unwell for the last couple of weeks and haven't really been able to look at this. I will try to look in more detail in the next couple of days, I would like to see an additional piece of logic:
* If the changelog placeholder is detected, emit a warning of deprecation before falling back to the old behaviour implemented
* I don't want to put the placeholder into the default changelog template, but I'm ok with a "legacy" changelog template that has one; or a switch to add it to the content.

I appreciate that might not be the most clear explanation, will come back to this in the near future

roggervalf:
> Hi @roggervalf - apologies, I've been quite unwell for the last couple of weeks and haven't really been able to look at this. I will try to look in more detail in the next couple of days, I would like to see an additional piece of logic:
> 
>     * If the changelog placeholder is detected, emit a warning of deprecation before falling back to the old behaviour implemented
> 
>     * I don't want to put the placeholder into the default changelog template, but I'm ok with a "legacy" changelog template that has one; or a switch to add it to the content.
> 
> 
> I appreciate that might not be the most clear explanation, will come back to this in the near future

thank you @bernardcooke53, no worries, take your time

semantic_release/version/algorithm.py
```diff
@@ -27,7 +27,7 @@
 
 
 def tags_and_versions(
-    tags: Iterable[Tag], translator: VersionTranslator
+    tags: Iterable[Tag], translator: VersionTranslator, only_last_release: bool = False
```
bernardcooke53, line 30: I don't think it's necessary to add this, it's not the purpose of `tags_and_versions` at all. There's only one location where you intend to set this to `True`, I don't see why you can't just add the `[0]` there?

semantic_release/cli/commands/changelog.py
```diff
@@ -62,10 +52,55 @@ def changelog(ctx: click.Context, release_tag: str | None = None) -> None:
...
-        changelog_text = render_default_changelog_file(env)
-        with open(str(changelog_file), "w+", encoding="utf-8") as f:
-            f.write(changelog_text)
+        with open(str(changelog_file), "a+", encoding="utf-8") as f:
+            f.seek(0)
+            existed_changelog_text = f.read()
+            f.truncate()
+            existed_changelog_sections = existed_changelog_text.split(CHANGELOG_PLACEHOLDER)
+            if len(existed_changelog_sections) > 1:
+                log.warning("Deprecation: Placeholder is detected, only new changes will be added.")
```
bernardcooke53, line 61: :+1: for warning, but it would be good to include a link to [Migrating an Existing Changelog](https://python-semantic-release.readthedocs.io/en/latest/changelog_templates.html#changelog-templates-migrating-existing-changelog)

semantic_release/cli/commands/changelog.py
```diff
@@ -62,10 +52,55 @@ def changelog(ctx: click.Context, release_tag: str | None = None) -> None:
             )
             ctx.exit(0)
 
-        changelog_text = render_default_changelog_file(env)
-        with open(str(changelog_file), "w+", encoding="utf-8") as f:
-            f.write(changelog_text)
+        with open(str(changelog_file), "a+", encoding="utf-8") as f:
+            f.seek(0)
+            existed_changelog_text = f.read()
+            f.truncate()
```
bernardcooke53, line 58: This is error-prone - if there's an exception later on then we will have truncated the changelog without writing any text back to it

semantic_release/cli/commands/changelog.py
```diff
@@ -62,10 +52,55 @@ def changelog(ctx: click.Context, release_tag: str | None = None) -> None:
...
+            existed_changelog_sections = existed_changelog_text.split(CHANGELOG_PLACEHOLDER)
+            if len(existed_changelog_sections) > 1:
```
bernardcooke53, line 60: I think it's a much simpler check to have `f"{CHANGELOG_PLACEHOLDER}\n" in f.readlines()` - avoids trying to split. The text could (by chance) by in a commit message which would lead to issues.

You can always do

```python3
existing_changelog_text = f.read()
f.seek(0)
lines = f.readlines()
```
to get both, or even
```python3
lines = f.readlines()
existing_changelog_text = "".join(lines)
```

semantic_release/cli/commands/changelog.py
```diff
@@ -62,10 +52,55 @@ def changelog(ctx: click.Context, release_tag: str | None = None) -> None:
             )
             ctx.exit(0)
 
-        changelog_text = render_default_changelog_file(env)
-        with open(str(changelog_file), "w+", encoding="utf-8") as f:
-            f.write(changelog_text)
+        with open(str(changelog_file), "a+", encoding="utf-8") as f:
```
bernardcooke53, line 55: why open in append mode?

semantic_release/cli/commands/changelog.py
```diff
@@ -62,10 +52,55 @@ def changelog(ctx: click.Context, release_tag: str | None = None) -> None:
...
+            f.seek(0)
+            existed_changelog_text = f.read()
+            f.truncate()
+            existed_changelog_sections = existed_changelog_text.split(CHANGELOG_PLACEHOLDER)
+            if len(existed_changelog_sections) > 1:
+                log.warning("Deprecation: Placeholder is detected, only new changes will be added.")
+                rh = ReleaseHistory.from_git_history(
```
bernardcooke53, line 62: We always iterate in reverse order over the commits, which means the latest release is the first one inserted into `ReleaseHistory.released`.

You can get the last release by doing:

```python3
full_rh = ReleaseHistory.from_git_history(<the existing params>)
latest_release_version = next(iter(full_rh.unreleased), None)
latest_release = full_rh.released.get(latest_release_version, {})

rh = ReleaseHistory(unreleased=full_rh.unreleased, released=latest_release)
```

This works because `dict`s remember their insertion order so `next` will find the oldest-inserted key

semantic_release/cli/commands/changelog.py
bernardcooke53, line 1: General comment - there's no testing for this functionality, would be nice to see some tests added for it

bernardcooke53 (commented):
I have left some comments - thank you for putting the code together!

I would want to see some changes before merging this, but am happy to help or to draft those up and get your review on them, as you prefer

semantic_release/cli/commands/changelog.py
roggervalf: hey @bernardcooke53, because I'm reading and writting

semantic_release/cli/commands/changelog.py
roggervalf: I think I need some help on making a test for this

semantic_release/cli/commands/changelog.py
roggervalf: with these suggestions I'm not sure how to keep the existing changelog, as I don't want to override previous changelogs, only add the latest changelog

PR #674, opened 2023-08-11
chore(deps): remove obsolete dataclasses dependency

afuetterer:
Hi there,

I think if python-semantic-release officially supports Python >= 3.7, you can get rid of the explicit `dataclasses` dependency (python_version < '3.7.0').  Is that right?

Feel free to close, if you still want to support older versions.


bernardcooke53 (approved):
:+1: agreed

PR #675, opened 2023-08-11
feat: infer gitea api_domain from domain, if unspecified.

lukester1975:
Hello again

For a gitea repo how about implying the `api_domain` (when not specified) from `domain`? Saves a bit more config.

Thanks

PS how about inferring `domain` itself (when not provided) from pyproject.toml `project.urls.repository` (if provided)? I had a quick look but given the code drops everything but `[tool.semantic_release]` from pyproject.toml immediately after parsing it, the change seemed like more than a one-liner. Happy to have a look if you think it's sensible.

### DIFF
--- a/semantic_release/hvcs/gitea.py
+++ b/semantic_release/hvcs/gitea.py
@@ -8,6 +8,7 @@ import mimetypes
 import os
 
 from requests import HTTPError
+from urllib3.util.url import Url, parse_url
 
 from semantic_release.helpers import logged_function
 from semantic_release.hvcs._base import HvcsBase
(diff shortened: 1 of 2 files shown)


bernardcooke53 (commented):
The change looks good @lukester1975 - would you mind adding a change to the docs explaining that this can be inferred for Gitea servers?

bernardcooke53:
Also re: your PS - as you say, I'm sure it's more than a one-line change. I think it's a nice idea, though, as long as the configuration parsing doesn't get super complex as a result

lukester1975:
I think I did look but couldn't see any existing documentation for `domain` and `api_domain` so thought maybe just the changelog would suffice?

lukester1975:
> Also re: your PS - as you say, I'm sure it's more than a one-line change. I think it's a nice idea, though, as long as the configuration parsing doesn't get super complex as a result

Struck me there's a simpler (almost 1 line) approach: https://github.com/lukester1975/python-semantic-release/commit/907840f66f72d5f6b599cbb1fc02144664847fe9

That said I don't think it's correct if the http access is via a subpath.

E.g. remote_url: `ssh://gitea@my.server/org/repo-name`
http url: `https://my.server/gitea/org/repo-name`

No way to infer that `/gitea` subpath component of the http path. But then the existing config option is available...

If you think this is a non-starter I'll see if I can find time to dig out the url from pyproject.toml.

Thanks

bernardcooke53:
You're right, the options should be under the [remote](https://python-semantic-release.readthedocs.io/en/latest/configuration.html#tool-semantic-release-remote) table but aren't yet.

Was thinking through the relative tradeoff between being simple and explicit (maybe causing duplication) and convenience having things auto-detected.
Could be sensible to do the following:
Try to derive api_domain from domain using some sensible defaults, if the config option is missing - those would be `api.{domain}` for GitHub, `{domain}/api/v4` for Gitlab and `{domain}/api/v1` for Gitea. In addition, we could try to detect domain from the following locations in order:

1. the `remote.domain` config option
2. `git config --get remote.{name}.url`, where `{name}` is the `remote.name` config option
3. `project.urls.repository`

That way we'd have a sensible set of locations and defaults, but we're somewhat consistent in how things are detected and when you have to specify things. Not suggesting all that has to go into a PR here, but keen to hear what you think :slightly_smiling_face:

bernardcooke53 (approved):

lukester1975:
Probably put 3 before 2 due to the ssh/https issue and subpaths?

PR #699, opened 2023-09-04
Update minimum version pin of GitPython to fix vulnerability.

rjdbcm:
Update minimum version of GitPython to 3.1.34 to fix resource leak issue introduced via a path traversal vulnerability in versions prior to 3.1.34.


rjdbcm:
I made this in error. Vulnerability is not mitigated in 3.1.34.

PR #708, opened 2023-09-14
docs: fix typos

afuetterer:
Hi, I think these might all be typos, right?

I was unsure in this line:
https://github.com/python-semantic-release/python-semantic-release/blob/3abfb7ac216b9ad439de24fda60eca84038e850e/tests/unit/semantic_release/version/test_declaration.py#L83

Is this a typo or some sort of fuzzy matching?
`r"(?P<vrsion>.*)"`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN set -ex; \
     apt-get install -y --no-install-recommends \
     git-lfs
 
-#install backported stable vesion of git, which supports ssh signing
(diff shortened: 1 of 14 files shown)


bernardcooke53 (approved):
Thank you - this is really great!

bernardcooke53:
> I was unsure in this line:

That's actually deliberate, to test for typos ironically :sweat_smile:

PR #732, opened 2023-10-06
fix: Error when running build command on windows systems

JCHacking:
On Windows systems, if the shell you get is cmd instead of executing the command it opens a shell
This is because instead of telling it cmd /c COMMAND it is sent cmd -c COMMAND

### DIFF
--- a/semantic_release/cli/commands/version.py
+++ b/semantic_release/cli/commands/version.py
@@ -102,7 +102,7 @@ def shell(cmd: str, *, check: bool = True) -> subprocess.CompletedProcess:
     if not shell:
         raise TypeError("'shell' is None")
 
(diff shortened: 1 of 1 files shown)


JCHacking:
Can any maintainer check if this is ok or do I need some more action on the PR? This is the second one I've opened on GitHub and I don't know if it requires any further action on my part.

bernardcooke53:
Hey @JCHacking - thank you for the PR, nothing else needed from your side!

JCHacking:
@bernardcooke53 When will a new version with this fix be released?

bernardcooke53:
@JCHacking this should be part of v8.1.2 now

PR #741, opened 2023-10-24
feat!(action): use composite action for semantic release (#692)

bernardcooke53:


bernardcooke53:
@zckv I have moved the changes from #692 over here.

I'm thinking I want to add the following to this PR (based off of the breakage in the previous one):

* Document additional requirements for workflows now that the action isn't Dockerized
  * e.g. add a `setup-python` step
* Add a workflow to test the GitHub action on runners using each of `ubuntu-latest`, `macos-latest` and `windows-latest`
  * I've been playing about with some usage [here](https://github.com/bernardcooke53/psr-692) which I think could be a starting point
* The entrypoint `action.sh` will need to be re-worked to be cross-platform, because when I ran this on `windows-latest` it fell over quite quickly - the easiest thing will probably be to just make it a Python script instead of shell.

Will make PRs against this branch for each of those, obviously no obligation but if you still want to work on the change would you mind doing the same please? That way it can all be merged at once :pray:

eikowagenknecht:
Would really love to see this PR go through because without it, projects that use a newer Python than 3.10 can't use the action at all. Pinning to 8.3.0 for now as well.

codejedi365:
> Would really love to see this PR go through because without it, projects that use a newer Python than 3.10 can't use the action at all. Pinning to 8.3.0 for now as well.

@eikowagenknecht, although this composite action has not been merged yet. #914 has and it bumped the Docker action up to Python:3.12-bookworm as the base environment which will allow building of projects that use a minimum version of `3.11` & `3.12`.

zckv:
Hello 👋 

Wouldn't it be simpler to create a separate action repository ? This solution would avoid a "breaking" change on this project for something that is not related to the PSR python code.

codejedi365:
> Wouldn't it be simpler to create a separate action repository ?

No, I don't believe so. It sounds nice up front but we would then have to keep both projects in sync. This action change would still be a breaking change which would then cause a version number drift of the version numbers of the project which likely confuses the configurations used and the users. This may not matter as the proper version is a set dependency in the action. 

I plan to wrap it in with the next breaking change for the project. The release will have a full migration instructions and such with it. Maybe at the breaking change it's moved to another repository which doesn't change all that much besides the version it points to upon this repository update.

eikowagenknecht:
> @eikowagenknecht, although this composite action has not been merged yet. #914 has and it bumped the Docker action up to Python:3.12-bookworm as the base environment which will allow building of projects that use a minimum version of `3.11` & `3.12`.

Thanks, working fine for 3.12 now.

codejedi365:
No update at this time however it looks like we will have a number of changes for `v10` and this will be included.

Significant migration docs will be needed and any additional changes this PR didn't have completed yet.

codejedi365:
Haven't gotten back around to reviewing and integrating this yet for v10

action.yml
```diff
@@ -21,7 +21,7 @@ inputs:
     required: false
   git_committer_email:
     description: "The email address for the “committer” field"
-    default: "github-actions@github.com"
+    default: "action@github.com"
     required: false
   ssh_public_signing_key:
     description: "The ssh public key used to sign commits"
```
suzhenskk, line 27: 123
```suggestion
    description: "The ssh public key used to sign commits"
```

codejedi365:
Still in backlog

PR #747, opened 2023-11-06
docs(migration): fix comments about publish command

codejedi365:
## Purpose

Update the migration documentation for clarity of what `publish` command does now as opposed to before.

## Rationale

This section was a great write up to explain how I can go about migrating from `v7` to `v8`, however I was very confused about the order of operations when I reviewed the version and publish code.  I corrected the steps and alignment to each command in this PR after reviewing the code segments.  This adjustment lines up with how `version` can/will run a new build and the second statement about removing the wrapped `twine` functionality.


bernardcooke53 (approved):
Thanks @codejedi365 - agreed this is clearer

PR #752, opened 2023-11-10
feat(cmd-version): add `--tag/--no-tag` option to version command

codejedi365:
## Purpose

- Provide the user the ability to turn off the application of tags to the repository.
- Provide the ability to prevent pushing commits when only using tags to indicate versions 

## Rationale

I need the ability to run `semantic-release` to apply versions across the file base prior to the build process and not have to manually remove the tag after it is executed. It fits in to the use of `--no-commit` but its not something I want to implement myself to commit-parse and apply to files when `semantic-release` already can provide this.  The reason for this is to ensure that during my pipeline builds the actual release artifact during the build stage and passes it along to be tested rather than re-building an artifact at the end of the pipeline (which then I'm releasing an artifact that isn't actually tested).

I also had to adjust the push mechanism because I do not intend to commit the generated files (like the changelog or the modified version files to the repository) If I use `--no-commit`, it won't push the tag and I do want the tag to be published.

## How I tested it

Wrote 2 integration tests that validate that versions are stamped into the code without creating a commit or a tag. The second test ensured that the tag was pushed but no commit was pushed.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -213,12 +213,25 @@ For example, assuming a project is currently at version 1.2.3::
 Whether or not to perform a ``git commit`` on modifications to source files made by ``semantic-release`` during this
 command invocation, and to run ``git tag`` on this new commit with a tag corresponding to the new version.
 
-If ``--no-commit`` is supplied, a number of other options are also disabled; please see below.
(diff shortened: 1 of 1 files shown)


bernardcooke53 (approved):
Looks good to me, the workflow you described makes sense and this seems well suited to it- thank you!

PR #760, opened 2023-12-06
test(commandline-main): prevent git gpgsign config from failing tests

codejedi365:
## Purpose

Fix issues when running command line tests when the global git configuration includes `commit.gpgsign = true`.

## Rationale

As my environment has the global configuration setting `commit.gpgsign` set to `true`, the scenario tests inherit this default setting.  When the scenario tests execute a commit, this setting applies. It was properly disabled in scenario tests but was not disabled in the fixture used by this `main` command line test.  This change replicates the scenario `git_repo.py` fixtures config modifications.


bernardcooke53 (approved):
👍 thank you!

PR #764, opened 2023-12-07
fix(cmd-version): handle committing of git-ignored file gracefully

codejedi365:
## Purpose

Fix a TypeError when one of the modified files during `version` command adds a file that matches a `.gitignore` pattern.

## Rationale

I personally don't commit a change to the version number in my files when building my distribution. To enable my software to know its version number I write it to a file that is imported `_version.py`. This file is generated by the build system `setuptools_scm` and it makes it easy to have this file git-ignored.  When semantic-release executed the update to the version file, it tried to add it to the index and git rejects with an exit code unless the `--force` command is provided.  My recommendation (this implementation) is to catch this error gracefully and print a warning to the developer in case they did not desire it to be ignored.

## How I tested

I modified the current testing infrastructure to instead of using `__init__.py` as the version holder to instead use `_version.py` that can be our concept for an ignored file. I figured this helps the tests evenly hit more failure possibilities as we already have a file `pyproject.toml` that is being committed with a version number change.  This way we have one ignored and one committed for all tests.

### DIFF
--- a/semantic_release/cli/commands/version.py
+++ b/semantic_release/cli/commands/version.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, ContextManager, Iterable
 
 import click
 import shellingham  # type: ignore[import]
+from git.exc import GitCommandError
 
 from semantic_release.changelog import ReleaseHistory, environment, recursive_render
 from semantic_release.changelog.context import make_changelog_context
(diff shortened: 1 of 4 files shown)


bernardcooke53 (approved):
🚀 nice, I agree with you we get more mileage out of an ignored `_version.py` as well

PR #765, opened 2023-12-07
fix(config): gracefully fail when repo is in a detached HEAD state

codejedi365:
## Purpose

Gracefully handle a TypeError when repository is in a detached HEAD state as occurs naturally in GitLab CI/CD environments.

## Rationale

As I was setting up this library to run within GitLab CI/CD, I ran into an unhandled TypeError because of the detached HEAD state that GitLab's Runner Helper image sets up the environment by default.  This is problematic due to the branch detection mechanism used in `python-semantic-release`. I opted to have the code gracefully catch this error and display a rationale to the user.  

I have devised a way to manually resolve this problem in GitLab in the step before `semantic-release` is called but have not implemented an internal correction method (match the detached HEAD SHA to an equivalent branch and checkout). At the current time it will be up to the developer to ensure the state is correct.

## How I tested

Outside of the manual integration tests I performed on GitLab, I added a test that purposefully mimics a detached HEAD state like GitLab does.

### DIFF
--- a/semantic_release/cli/config.py
+++ b/semantic_release/cli/config.py
@@ -257,8 +257,18 @@ class RuntimeContext:
         ##
         # credentials masking for logging
         masker = MaskingFilter(_use_named_masks=raw.logging_use_named_masks)
+
+        try:
+            active_branch = repo.active_branch.name
+        except TypeError as err:
+            raise NotAReleaseBranch(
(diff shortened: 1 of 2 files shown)


bernardcooke53 (approved):
👍

PR #766, opened 2023-12-08
docs(configuration): adjust wording and improve clarity

codejedi365:
## Purpose

Update mis-spelled words and adjust wording for clarity

## Rationale

I adjusted the environment variable section a bit because if people don't read and just look for code blocks then we have an example in there that is not recommended and its not easy to tell its not recommended.  I recommend we just eliminate that example completely but state it clearly that it should not be done that way.

Other items were just clarity of language. Feel free to pass feedback if my wording is not clear either.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -4,9 +4,9 @@ Configuration
 =============
 
 Configuration is read from a file which can be specified using the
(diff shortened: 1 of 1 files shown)


bernardcooke53 (approved):
🥇

Issue #734, opened 2023-10-13
How can I use it purely locally


Deran80:
Hello,

how can I use python-semantic-release without a remote repository?
Or with a 'local' remote repository / directory (git init --bare)?
Due to company guidelines I'm not allowed to upload the code to github.

If I dont't add a remote repository, I get the error 'ValueError: Remote named 'origin' didn't exist'
If I add a local directory as a remote repository, I get the error 'ValueError: Cannot parse ...' because it doesn't match an url pattern.
Although I get these errors, python-semantic-release increments the version number, but it doesn't create a changelog.
Git can push to the directory.

Is there a trick how I can get this working?

Thanks in advance.

codejedi365:
This is an interesting question. There might be something you can do. Try

```sh
semantic-release version --no-push
```

codejedi365:
@Deran80, I got a concept working that will support your offline use. I am still working on the testing but soon I'll have a PR and a release that you can use.

codejedi365:
This is actually very close to resolution after some refactors of the base interface classes. Still requires testing and updates to the latest interface. 

Unfortunately lower on the priority compared to some current and recent issues.

codejedi365:
Same update as previously, not much has moved forward as I have been very busy.

codejedi365:
Still need to get back to this as a lot has changed in recent months. Hopefully soon.

codejedi365:
Still in backlog

codejedi365:
Still in backlog

codejedi365:
Still in backlog

codejedi365:
Still under development but it has stalled a bit.

codejedi365:
Still in backlog

PR #771, opened 2023-12-14
feat(utils): expand parsable valid git remote url formats

codejedi365:
## Purpose

Expand the compatibility to as many variants of valid git remote urls as git allows

## Rationale

In preparation to facilitate the request of #734, first we need to properly parse all valid git urls.  I found a stackoverflow reference that listed many of the possible urls and I built a unit test to match a set of them.  During implementation, I opted to coerce each input into an URL-parse compatible format which then could be easily broken up into the valid parts.

## How I tested

Build a paramaterized unit test with all the expected valid tests.

## How to verify

```sh
git fetch origin pull/771/head:pr-771

git checkout pr-771~1 --detach

# execute tests, 10 should fail with valid urls
pytest tests/unit/semantic_release/test_helpers.py

git merge --ff-only pr-771

# run pytest again
pytest tests/unit/semantic_release/test_helpers.py
```

### DIFF
--- a/semantic_release/helpers.py
+++ b/semantic_release/helpers.py
@@ -3,6 +3,7 @@ import logging
 import re
 import string
 from functools import lru_cache, wraps
+from pathlib import PurePosixPath
 from typing import Any, Callable, NamedTuple, TypeVar
 from urllib.parse import urlsplit
 
@@ -73,7 +74,7 @@ def dynamic_import(import_path: str) -> Any:
 
 
 class ParsedGitUrl(NamedTuple):
-    """Container for the elements parsed from a git URL using GIT_URL_REGEX"""
+    """Container for the elements parsed from a git URL"""
 
     scheme: str
     netloc: str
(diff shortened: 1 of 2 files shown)


codejedi365:
@bernardcooke53, @relekang, would you mind doing a review?

tests/unit/semantic_release/test_helpers.py
```diff
@@ -0,0 +1,82 @@
...
+@pytest.mark.xfail(raises=ValueError)
+@pytest.mark.parametrize('url', [
+    "icmp://git",
+    "abcdefghijklmnop.git",
+    "../relative/path/to/repo.git",
+    "http://domain/project.git"
+])
+def test_parse_invalid_git_urls(url: str):
+    """Test that an invalid git remote url throws a ValueError."""
+    assert parse_git_url(url)
```
bernardcooke53, line 82: I don't think this is the intended usage for `xfail` - I'd suggest using `pytest.raises` instead:

```suggestion
@pytest.mark.parametrize('url', [
    "icmp://git",
    "abcdefghijklmnop.git",
    "../relative/path/to/repo.git",
    "http://domain/project.git"
])
def test_parse_invalid_git_urls(url: str):
    """Test that an invalid git remote url throws a ValueError."""
    with pytest.raises(ValueError):
        parse_git_url(url)
```

semantic_release/helpers.py
```diff
@@ -73,67 +74,94 @@ def dynamic_import(import_path: str) -> Any:
...
         raise ValueError(f"Cannot parse {url!r}")
 
-    repo_name = m.group("repo_name")
-    repo_name = repo_name[:-4] if repo_name.endswith(".git") else repo_name
+    # We have been able to parse the url with urlsplit,
+    # so it's a (file|git|ssh|https?)://... structure
+    # but we aren't validating the protocol scheme as its not our business
+
+    # use PosixPath to normalize the path & then separate out the namespace & repo_name
+    namespace, _, name = str(PosixPath(urllib_split.path)).lstrip("/").rpartition("/")
```
bernardcooke53, line 145: Given we're just using this for string normalisation I think it's a bit clearer to use [`PurePosixPath`](https://docs.python.org/3.7/library/pathlib.html#pathlib.PurePosixPath), as `Pure*Path` classes don't make any syscalls

bernardcooke53 (changes requested):
In general looks good to me, just a couple of minor comments 🙂

semantic_release/helpers.py
codejedi365: I did not know that! I learned today! Thx!

tests/unit/semantic_release/test_helpers.py
codejedi365: wow, I just read more into the docs on this. I totally misunderstood the first time, `pytest.raises()` is definitely the correct way. I'm glad you pointed this out.

codejedi365:
@bernardcooke53, thanks for the comments, fixed!

semantic_release/helpers.py
bernardcooke53: I only found that difference out myself pretty recently 😁

bernardcooke53 (approved):
Looks good to me - thanks @codejedi365!

PR #772, opened 2023-12-17
fix(cli): gracefully output configuration validation errors

codejedi365:
## Purpose

Improve handling of exceptions to provide a useful message to user without stack trace

## Rationale

If a user puts in an invalid configuration setting, the configuration validator will throw a ValidationException that we do not catch gracefully. This causes python to dump a stack trace and hide the pretty ValidationError that pydantic provides by default.  

In order to set up the test appropriately, I found it useful to add a fixture that will modify the configuration file accordingly.  Meanwhile, I noticed that the implementation of example project was a bit fragile as exceptions are not handled well with generators or context managers unless explicitly using try/finally. With this modification I was able to align it more with the setup/teardown generator style fixtures as pytest recommends.

## How I tested

Created a command line test that modifies the configuration file to be incorrect before executing the `version` command. It evaluates the exit code and the stderr message to the user that there is a validation error and which component.

## How to verify

```sh
git fetch origin pull/772/head:pr/772

git checkout pr/772~1 --detach

# execute tests, should fail
pytest tests/ -k test_errors_when_config_file_invalid_configuration

git merge --ff-only pr/772

# run pytest again (looking for success)
pytest tests/ -k test_errors_when_config_file_invalid_configuration

# run entire test suite for regression testing
PYTHONHASHSEED=123456 pytest tests/
```

### DIFF
--- a/semantic_release/cli/commands/main.py
+++ b/semantic_release/cli/commands/main.py
@@ -7,6 +7,7 @@ import click
 from click.core import ParameterSource
 from git import InvalidGitRepositoryError
 from git.repo.base import Repo
+from pydantic import ValidationError
 from rich.console import Console
 from rich.logging import RichHandler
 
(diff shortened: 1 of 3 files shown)


codejedi365:
@bernardcooke53, @relekang, would you mind doing a review?

bernardcooke53 (approved):
LGTM, thanks @codejedi365 🎉

PR #773, opened 2023-12-17
test(infrastructure): refactor test fixtures & configuration for higher resiliency

codejedi365:
## Purpose

1. Refactor test infrastructure for easier extensibility and more accurate testing
2. Refactor to specify exact modifications to configurations within tests & fixtures.  This improves the maintainability of the test code from the source code's internal defaults instead of a hardcoded configuration which also is harder to change on the fly.
3. Enable the non-tox method of running pytest to still work with `xdist`

## Rationale

Regarding item 1, I was able to fix the warning issue of too many files open by implementing a pytest setup/teardown in the generator implementation. This means `repo.close()` is automatically called via the `try/finally` which is necessary in case a test case fails, it will still execute the teardown-cleanup step. This should also help with the cascading errors that have occurred when one test fails, they all fail from then on.

Regarding item 2, I removed the base configuration we were using to make sure the tests started with the bare minimum (ie the internal defaults) and then we can modify from then when appropriate.  One such case, is all of the repo scenarios, which ended up having deliberately non-angular commits but the base config only had angular commit options specified and the angular parser specified.  In this new refactor, the settings map to the repo's commit structure .  This lays the ground work for a few changelog tests that were failing due to mix-matched configurations.

I tend to not run `tox` as it runs every version of python and takes forever while I'm developing, therefore, I moved the HASHSEED configuration to pytest's config instead which also is effectively inherited in `tox` as well.

## How I tested

Ran the test suite via pytest and tox to ensure viability. I also evaluated the changelog outputs manually on non-angular scenarios created by the fixtures (will be in a future PR) and these changes really helped.

## How to verify

```sh
git fetch origin pull/773/head:pr-773
git checkout pr-773

# run pytest to see success
pytest -vv tests
```


codejedi365:
@bernardcooke53, this PR is ready for review when you have time, thanks

tests/fixtures/git_repo.py
```diff
@@ -31,17 +48,18 @@ def example_git_https_url():
...
+    request: pytest.FixtureRequest,
+    example_project: ExProjectDir
+) -> Generator[RepoInitFn, None, None]:
+    repos: list[Repo] = []
+
+    def git_repo() -> Repo:
         repo = Repo.init(example_project.resolve())
+
+        # store the repo so we can close it later
+        repos.append(repo)
```
bernardcooke53, line 61: 👍 this is neat, though it will keep a reference to `Repo` instances for longer than they used to exist. I don't see that being a problem, but if it ever is then a `weakref.WeakSet` should help

bernardcooke53 (approved):
This looks great - thanks a lot @codejedi365 🚀

PR #774, opened 2023-12-17
refactor(config): enable default environment token per hvcs

codejedi365:
## Purpose

Update the configuration parser to optionally default to the VCS specific token

## Rationale

In preparation to facilitate the request of #734, I need to find a way to solve the missing token when the `remote.type` is `none`.  This led me to determining that it would be helpful if we looked for the default token names for each of the VCS. This refactor enables the configuration to be aware of each VCS's default.  

Current downside is that `Gitea` does not provide a pre-authed CI-like token based on [this discussion](https://github.com/go-gitea/gitea/issues/20394).  I just guessed to use `GITEA_TOKEN`. I could potentially have it be empty instead, which would force the user to provide it.

Consideration also should be made for `GitLab` as GitLab's default token `CI_JOB_TOKEN` has permissions problems with `python-semantic-release`.  The fact that `python-semantic-release` uses `git push` both commits and tags are both unauthorized actions by a `CI_JOB_TOKEN`.  This means users are forced to provide a PAT as that is how we mark versions. What the users define as that token name is up to them. My guess would be to use `GITLAB_TOKEN` but maybe we need to force them to be specific here also. Curious on your thoughts.

Separately, there is still some investigation that I'm doing here to hopefully find a better implementation for GitLab.

One of the other issues (besides for `none`) that I have with the current implementation is that if the user does not know to specify a token value or forgets, we default to `GH_TOKEN` when I have specified `remote.type = 'gitlab'`.

## How I tested

Build a paramaterized unit test with all the expected valid & invalid tests.

## How to verify

```sh
git fetch origin pull/774/head:pr-774

git checkout pr-774~1 --detach

# execute tests, 2 should fail (non-github)
pytest tests/unit/semantic_release/cli/test_config.py

git merge --ff-only pr-774

pytest tests/unit/semantic_release/cli/test_config.py
```

### DIFF
--- a/semantic_release/cli/config.py
+++ b/semantic_release/cli/config.py
@@ -6,14 +6,15 @@ import re
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import Any, ClassVar, Dict, List, Optional, Tuple, Union
+from typing import Any, ClassVar, Dict, List, Optional, Tuple, Type, Union
 
 from git import Actor
 from git.repo.base import Repo
 from jinja2 import Environment
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
 from typing_extensions import Literal
 
+from semantic_release import hvcs
 from semantic_release.changelog import environment
 from semantic_release.cli.const import DEFAULT_CONFIG_FILE
 from semantic_release.cli.masking_filter import MaskingFilter
@@ -29,7 +30,6 @@ from semantic_release.commit_parser import (
 from semantic_release.const import COMMIT_MESSAGE, DEFAULT_COMMIT_AUTHOR, SEMVER_REGEX
 from semantic_release.errors import InvalidConfiguration, NotAReleaseBranch
 from semantic_release.helpers import dynamic_import
-from semantic_release.hvcs import Gitea, Github, Gitlab, HvcsBase
 from semantic_release.version import VersionTranslator
 from semantic_release.version.declaration import (
     PatternVersionDeclaration,
@@ -44,6 +44,21 @@ class HvcsClient(str, Enum):
     GITHUB = "github"
     GITLAB = "gitlab"
     GITEA = "gitea"
+
+
+_known_commit_parsers = {
+    "angular": AngularCommitParser,
+    "emoji": EmojiCommitParser,
+    "scipy": ScipyCommitParser,
+    "tag": TagCommitParser,
+}
+
+
+_known_hvcs: Dict[HvcsClient, Type[hvcs.HvcsBase]] = {
+    HvcsClient.GITHUB: hvcs.Github,
+    HvcsClient.GITLAB: hvcs.Gitlab,
+    HvcsClient.GITEA: hvcs.Gitea,
+}
 
 
 class EnvConfigVar(BaseModel):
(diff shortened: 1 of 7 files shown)


semantic_release/cli/config.py
```diff
@@ -29,7 +29,7 @@
 from semantic_release.const import COMMIT_MESSAGE, DEFAULT_COMMIT_AUTHOR, SEMVER_REGEX
 from semantic_release.errors import InvalidConfiguration, NotAReleaseBranch
 from semantic_release.helpers import dynamic_import
-from semantic_release.hvcs import Gitea, Github, Gitlab, HvcsBase
+from semantic_release.hvcs import *
```
bernardcooke53, line 32: I'm not a fan of glob imports unless they're for a specific purpose, generally I find it clearer to either import the name directly or import the namespace and then access stuff as attributes

semantic_release/cli/config.py
```diff
@@ -90,13 +105,22 @@ class BranchConfig(BaseModel):
...
     name: str = "origin"
-    token: MaybeFromEnv = EnvConfigVar(env="GH_TOKEN")
+    token: MaybeFromEnv = ""
     url: Optional[MaybeFromEnv] = None
     type: HvcsClient = HvcsClient.GITHUB
     domain: Optional[str] = None
     api_domain: Optional[str] = None
     ignore_token_for_push: bool = False
 
+    @model_validator(mode="after")
```
bernardcooke53, line 115: 👍 good approach imo

tests/unit/semantic_release/cli/test_config.py
```diff
@@ -1,15 +1,37 @@
...
+    from typing import Any
+
+
+@pytest.mark.parametrize("remote_config, expected_token", [
+    ({ "type": HvcsClient.GITHUB.value }, EnvConfigVar(env="GH_TOKEN")),
+    ({ "type": HvcsClient.GITLAB.value }, EnvConfigVar(env="GITLAB_TOKEN")),
+    ({ "type": HvcsClient.GITEA.value }, EnvConfigVar(env="GITEA_TOKEN")),
+    ({}, EnvConfigVar(env="GH_TOKEN")),  # default not provided -> means Github
+    pytest.param({ "type": "nonexistent" }, EnvConfigVar(env=""), marks=pytest.mark.xfail),
+])
```
bernardcooke53, line 28: I think this would be better as a separate test - `xfail`-ing here could hypothetically mask some other unexpected error that only this test case would produce

bernardcooke53 (changes requested):
Looks solid, just a couple of minor comments - thanks @codejedi365 🚀

tests/unit/semantic_release/cli/test_config.py
codejedi365: After realizing I was using xfail incorrectly, adjusted accordingly to make another test. That other test however does just fail at the common configuration validation check prior to getting to the default token state.

bernardcooke53: That's ok - I think that gives some coverage to the code path when the remote type isn't one of the three supported ones, anyway

bernardcooke53 (approved):
Thanks @codejedi365! 🚀

PR #782, opened 2023-12-26
fix(config): set commit parser opt defaults based on parser choice

codejedi365:
## Purpose

In preparation for a simpler configuration with dependent default options and a per parser-type config generator, this PR pulls the defaults from the associated commit parser value if not provided.

## Rationale

Simplifying the amount of options that the user must specify especially for non-angular parsers. The defaults for non-angular parsers are only available in the documentation. This way they just have to remove the options value completely after a `generate-config` command and they can just use the built-in defaults.  I have plans to extend the `generate-config` functionality in the future to allow these defaults to be generated instead and this PR would need to be accepted first.

## How I tested

Added additional unit tests which validated that the default commit parser options were loaded for the proper commit parser when no parser options were provided.

## How to verify

```sh
git fetch origin pull/782/head:pr-782

git checkout pr-782~2 --detach

# execute tests
pytest tests/unit/semantic_release/cli/test_config.py

git merge --ff-only pr-782

pytest tests/unit/semantic_release/cli/test_config.py
```

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -207,27 +207,61 @@ or transformation.
 
 For more information, see :ref:`commit-parsing-parser-options`.
 
-The default values are the defaults for :ref:`commit-parser-angular`
+The default value for this setting depends on what you specify as
+:ref:`commit_parser <config-commit-parser>`. The table below outlines
+the expections from ``commit_parser`` value to default options value.
 
-**Default:**
+==================  ==   =================================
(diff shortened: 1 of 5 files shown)


bernardcooke53 (approved):
This looks sensible to me - thanks @codejedi365 🎉

codejedi365:
@bernardcooke53, rebase completed, thanks for reviewing ahead of time!

PR #786, opened 2023-12-29
docs(configuration): change defaults definition of token default to table

codejedi365:
## Purpose

Improve readability of defaults relating to `remote.token`.

## Rationale

I forgot to update the docs with #774, and I was thinking this would be clearer than the paragraph of consecutive if-then statements in text.

## Evaluate

```sh
sphinx-autobuild --open-browser docs docs/_build/html
```

http://127.0.0.1:8000/configuration.html#token-dict-env-str

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -579,12 +579,16 @@ environment variable if you wish.
 
 
 The default value for this setting depends on what you specify as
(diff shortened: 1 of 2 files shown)


bernardcooke53 (approved):
👍 much clearer, thanks

PR #787, opened 2023-12-31
docs(contributing): add docs-build, testing conf, & build instructions

codejedi365:
## Purpose

Add additional testing and development instructions for an contributing developer

## Rationale

I was unfamiliar with the some of tools and the configuration of this project. This adjustment helps other developers not have to figure it out but instead have the commands on hand to know what to do.

## How I tested

I manually executed all the new instructions added here during my development. I ran a VSCode debugger on a pytest function and was able to break at a desired breakpoint in the test and in the code.  For the docs, the `_build` was important for the `sphinx-autobuild` watchdog server to not trigger a rebuild on the creation of the docs initially.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -29,19 +29,69 @@ that should result in a new release it will happen if the build is green.
 Development
 ~~~~~~~~~~~
 
-Install this module and the development dependencies::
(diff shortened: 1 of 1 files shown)


bernardcooke53 (approved):
LGTM

PR #788, opened 2024-01-02
test: refactor git repo fixtures

codejedi365:
> **Recommend** Reviewing this PR through the commit by commit view as opposed to the Files Changed view, it will be easier to understand as you can match my commit message with the changes made.  I apologize up front as this is a large amount of changes.

## Purpose

A phase 4 improvement to the repo scenario builders and test implementation for more readable and distinguishable results

## Rationale

I simplified much of the repetitive function calls in each of the repo builders to try and simplify adjustments in the future.  You will see some remnants that foreshadows my modifications for changelog validation that will make more sense in the next PR.  The new design basically provides a common Repo Definition that puts all the commit messages under the version value that it creates.  It reads a bit like a changelog where the commit is what creates the next version.  I was able to remove many of the hardcoded references using a generator and this definition. It will be easier to maintain since all the commit types are right next to one another as well.  Contrary to DRY, I did opt to not use a loop to generate and make versions but instead copy-paste the same code each time. I suspect I will have changes here soon as it will be harder to have logic that can be determined by a loop from the repo structure. It gets confusing where some must branch and others not.

You will also notice I created some fixture helper functions that execute common tasks in one go for the repo creation actions.  I opted to also drop out some of the file modifications as we can just use the change in `pyproject.toml` as something to commit when making a new version.

## How I tested

Primarily this is a testing refactor so all the current tests should stay the same just ensure that the fixtures set up the repos properly.  Similar to the caching, I would evaluate the pytest temporary environment after running the repo fixtures and run commands like `git log --graph --decorate --oneline --all` to check the git tree and then also review the local files manually.  I did piece this PR off from the changelog evaluation PR I've been working so I was validating that the resulting repo will generate the expected changelog then.  Unfortunately those tests are not included here.

## How to verify

Manually review the pytest temporary directory and evaluate the `pyproject.toml` that will match the latest version tag. It will also have the correct configurations for the repository as well.  I like walking through the debugger while having another window in the temporary/cache directory, to make sure the build of the repo is as intended.

Otherwise, its the CI that provides the proof that it still works as expected.


codejedi365:
@relekang, @bernardcooke53, pr ready for review for when you have time, thanks!

codejedi365:
@relekang, rebased and ready for review

relekang (approved):

PR #795, opened 2024-01-06
chore: modernize ruff configuration to work with ruff >= 0.2

afuetterer:
This PR proposes the following changes:

- update the ruff configuration to use e.g. `[tool.ruff.lint]` and `[tool.ruff.lint.flake8-quotes]`

Refs:
- https://docs.astral.sh/ruff/configuration/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -163,9 +163,17 @@ allow_incomplete_defs = true
 allow_untyped_calls = true
 
 [tool.ruff]
+line-length = 88
+target-version = "py38"
+force-exclude = true
+output-format = "grouped"
+show-fixes = true
+src = ["semantic_release", "tests"]
+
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -257,22 +265,15 @@ ignore = [
 ]
 
 external = ["V"]
-target-version = "py38"
-force-exclude = true
-line-length = 88
-output-format = "grouped"
 ignore-init-module-imports = true
-show-source = true
```
afuetterer, line 265: The ruff docs state that [show-source](https://docs.astral.sh/ruff/settings/#show-source) is deprecated.

You can get this behavior by setting output_format to full. What do you think?

afuetterer:
I fixed the issues @bernardcooke53.

Issue #663, opened 2023-07-26
automatic release to Gitlab


eitanief:
_Please describe your problem. Include links to any relevant pages or files._
On latest version documentation I see no automatic release guide for Gitlab, did you stop supporting this? It was previously supported

beatreichenbach:
You can find that information [here](https://python-semantic-release.readthedocs.io/en/latest/index.html#gitlab-gl-token).

codejedi365:
@eitanief, GitLab is definitely still supported, and we have a issue (#666) open to write a GitLab setup page but it isn't there yet. The information we have is scattered in current issues too. If you have a specific question, please let us know

Issue #768, opened 2023-12-11
insufficient_scope after first tag push to gitlab


damnmso:
## The problem

I added a bug report before but it either seems to be deleted or i can not find it anymore.
We have a jenkins pipeline that scans all 5 minutes for changes on our main/master branch in gitlab.
When a change is detected, we use semantic_versioning to generate the new version and push that tag to gitlab.
That works but after the tag has been created, the semantic_versioning aborts with a very wierd error:

```
[semantic_release.cli.commands.version]      version.py:553
                    ERROR version.version: 403:                                
                    insufficient_scope                                         
                    +--- Traceback (most recent call last) ----+               
                    | C:\Users\jenkinsSA\AppData\Local\Jenkins |               
                    | \.jenkins\workspace\usermgmtautomation_m |               
                    | aster\.pyenv-python\Lib\site-packages\gi |               
                    | tlab\exceptions.py:336 in wrapped_f      |               
                    |                                          |               
                    |   333         @functools.wraps(f)        |               
                    |   334         def wrapped_f(*args: Any,  |               
                    |   335             try:                   |               
                    | > 336                 return f(*args, ** |               
                    |   337             except GitlabHttpError |               
                    |   338                 raise error(e.erro |               
                    |   339                                    |   
```

unfortunately, a lot of code is "cut away" which might help a little to understand whats going on.
We have a "bug" in our pipeline that also pushed tags are recognized and the pipeline starts.
The second run then succeeds because semantic_versioning does not create a new version, therefore does not have to push tags.
But as mentioned: the git tag has been created. I dont understand why the plugin fails.

## Expected behavior
To not return an error, i mean it does what it should, it creates the tag, but it somehow crashes after that

## Environment

Please state which OS you are using and provide the output of the following commands:

```shell
Python 3.11.4
pip 23.1.2 from C:\Program Files\Python\Lib\site-packages\pip (python 3.11)
python-semantic-release==8.3.0

altgraph==0.17.4
annotated-types==0.6.0   
blinker==1.7.0           
certifi==2023.11.17      
cffi==1.16.0             
charset-normalizer==3.3.2
click==8.1.7             
colorama==0.4.6          
configparser==6.0.0      
coverage==7.3.2          
cryptography==41.0.5     
defusedxml==0.7.1
distlib==0.3.7
dotty-dict==1.3.1
Flask==3.0.0
from-root==1.3.0
gitdb==4.0.11
GitPython==3.1.40
idna==3.6
[...]

PR #797, opened 2024-01-06
feat: add a new type of hvcs for gitlab to support Gitlab Job Tokens

aborsu:
This supports GITLAB_CI exclusively via CI_JOB_TOKEN.

Switching from PAT to JOB-TOKEN is relatively easy as it  only requires switching the key used in the header  from PRIVATE-TOKEN to JOB-TOKEN respectively.

But the the original Gitlab hvcs class supports running outside of gitlab-ci and thus queries the API to get information that is available in the environment of the gitlab-ci job. 
Additionally, some of the queries used by the original Gitlab class are not allowed using a Job Token.
 cfr. [api-usage](https://python-gitlab.readthedocs.io/en/stable/api-usage.html)
``` python
# make an API request to create the gl.user object. This is not required but may be useful
# to validate your token authentication. Note that this will not work with job tokens.
gl.auth()
```

Supporting both running inside and outside Gitlab-CI within the same class is also possible, but we would have to make it explicit to avoid conflicts in the config resolution.

### DIFF
--- a/docs/automatic-releases/gitlab-ci.rst
+++ b/docs/automatic-releases/gitlab-ci.rst
@@ -1,0 +1,137 @@
+.. _github-actions:
+
+Setting up python-semantic-release on Gitlab CI
+====================================================
+
+Python Semantic Release can run on Gitlab CI using two mechanisms.
+The first one uses a personal access token and the second one uses a Job Token.
+
+Gitlab Ci `Job Token <https://docs.gitlab.com/ee/ci/jobs/ci_job_token.html>`_ are generated on the fly when a CI
+pipeline is triggered.
+The token has the same permissions to access the API as the user that caused the job to run, but with a subset of API
+routes available. While safer than a
+`Personal Access Token (PAT) <https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html>`_, it is not able
+to push new commits to the repository although discussions are ongoing on the subject (cfr `issue 389060
+<https://gitlab.com/gitlab-org/gitlab/-/issues/389060>`_).
+
+
+Using a personal access token (PAT)
+-------------------------------------
+
+Once you have `created a PAT <https://docs.gitlab.com/ee/user/profile/personal_access_tokens
+.html#create-a-personal-access-token>`_,  you can add it as a secret variable in your Gitlab CI project under the
+name GL_TOKEN.
+
+In order to use gitlab token you need to configure python semantic release by adding the following to your
+pyproject.toml file:
+
+.. code:: toml
+
+    [tool.semantic_release]
+    remote.type = "gitlab"
+
+Normally using  the following .gitlab-ci.yml should be enough to get you started:
+
+.. code:: yaml
+
+    stages:
+      - publish # Deploy new version of package to registry
+
+    # Official language image. Look for the different tagged releases at:
+    # https://hub.docker.com/r/library/python/tags/
+    image: python:latest
+
+    variables:
+      PIP_CACHE_DIR: $CI_PROJECT_DIR/.cache/pip # Set folder in working dir for cache
+
+    # Runs on commit in main branch that were not made by the semantic-release job
+    version_and_publish:
+      stage: publish
+      image: python:latest
+      variables:
+        GIT_DEPTH: 0
+      before_script:
+        - pip install python-semantic-release
+        - pip install twine
+      script:
+        - git checkout "$CI_COMMIT_REF_NAME"
+        - semantic-release version
+        - |
+          if [ "dist" ]; then
+            TWINE_PASSWORD=${CI_JOB_TOKEN} TWINE_USERNAME=gitlab-ci-token python -m twine upload --repository-url ${CI_API_V4_URL}/projects/${CI_PROJECT_ID}/packages/pypi dist/*
+          fi
+      cache:
+        paths:
+          - ${PIP_CACHE_DIR}
+      rules:
+        # Don't run on automatic commits
+        - if: $CI_COMMIT_AUTHOR =~ /semantic-release.*/
+          when: never
+        # Only run on main/master branch
+        - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
+
+Using a Job Token
+------------------
+In order to use gitlab-ci job token you need to configure python semantic release by adding the following to your
+pyproject.toml file:
+
+.. code:: toml
+
+    [tool.semantic_release]
+    remote.type = "gitlab-ci"
+    remote.ignore_token_for_push = "true"
+
+The following workflow is proposed to the reader.
+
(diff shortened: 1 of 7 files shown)


aborsu:
I'm still planning to write documentation for this, but I would appreciate a first feedback on the approach.
Thank you

aborsu:
This should resolve #663, please tell me if I'm forgetting something?

codejedi365:
@aborsu, I appreciate you taking the initiative on this. I'll take a look more in depth tomorrow. I would rather not cause confusion between a GitLab CI and a regular GitLab as it is the intent each HVCS is to be able to be run in the CI and locally.  

I had the same thought to have it work with the CI_JOB_TOKEN which means we have to modify the GitLab(*_token) call accordingly. Personally I was thinking of implementing the function in python-gitlab's cli.py which will derive the correct token based on the environment which I think is what we need but also a merge between using the current configuration parsing for environment variables.

The other problem we have is the call to `gl.auth()`, this did not work for me with a CI_JOB_TOKEN.  Were you able to get this to work?

Separately, when I tried to solve this problem previously, the tag publishing is the biggest limitation.  If you run `semantic-release version --push` then it will try to execute `git push --tags` to publish the commit & or tag to the repository.  The CI_JOB_TOKEN does not have enough scoped permissions for this. It can write to releases but it cannot write to the repository which is where the tags go.  Secondly, I don't think you can make a vcs release without a tag to link it too.  I have detailed this further in #768.  

Were you able to solve these problems?

aborsu:
@codejedi365 I can try to bring the two together later this week.
As I wrote in my first message gl.auth() will not work with a CI_JOB_TOKEN as written in the documentation of [python-gitlab api-usage](https://python-gitlab.readthedocs.io/en/stable/api-usage.html). 
This is because the CI_JOB_TOKEN has only access to a subset of the API, and on top of that there is no way to push using a CI_JOB_TOKEN cfr. [389060](https://gitlab.com/gitlab-org/gitlab/-/issues/389060).

Using the CI_JOB_TOKEN with [python-gitlab](https://python-gitlab.readthedocs.io/en/stable) is broken as many methods do not work. At the same time all the information is available in the environment variables. Basically, I would be fine putting a choice job_token or pat_token in a single class, the difficulty is the fact that the current Gitlab class works from outside the CI, mine doesn't so that would be a breaking change.

My ideal workflow (but only works on gitlab ultimate or premium), is that we have a Project Access Token as a guest on the project with a token that only allows it to push to master. It cannot use the API, it cannot push to another branch.
We use those credentials to push the commit and tag.

We then use the CI_JOB_TOKEN to create the release in HVCS.

aborsu:
I added the following doc

# Using a Job Token

In order to use gitlab-ci job token you need to configure python semantic release by adding the following to your
pyproject.toml file:

``` toml
    [tool.semantic_release]
    remote.type = "gitlab-ci"
    remote.ignore_token_for_push = "true"
```

The following workflow is proposed to the reader.

A [Project Access Token](https://docs.gitlab.com/ee/user/project/settings/project_access_tokens.html) is created in Gitlab and given the right to push to the main branch.
This token is added as a secret variable in your Gitlab CI project under the name `GL_PROJECT_TOKEN_NAME` and
`GL_PROJECT_TOKEN`. This token should only have write access to the repository, it does not require API access.
If you are using Gitlab Premium or Ultimate, you can make this token a guest to further restrain the token's scope,
and specify it as the sole person allowed to push to the master branch.
Otherwise, you will have to grant that project access token a sufficiently high access privilege that it can push to
the main branch.

The following .gitlab-ci.yml should be enough to get you started:
``` yaml
    stages:
      - publish # Deploy new version of package to registry
    # Official language image. Look for the different tagged releases at:
    # https://hub.docker.com/r/library/python/tags/
    image: python:latest
    variables:
      PIP_CACHE_DIR: $CI_PROJECT_DIR/.cache/pip # Set folder in working dir for cache
    # Runs on commit in main branch that were not made by the semantic-release job
    # Using GITLAB_USER_EMAIL in the GIT_COMMIT_AUTHOR will display the person who
    # triggered the job either by clicking the merge button or pushing to master
    # as the author of the commit.
    version_and_publish:
      stage: publish
      image: python:latest
      variables:
        GIT_DEPTH: 0
        GIT_COMMIT_AUTHOR: "$GL_PROJECT_TOKEN_NAME <$GITLAB_USER_EMAIL>"
      before_script:
        - pip install python-semantic-release
        - pip install twine
      script:
        - git checkout "$CI_COMMIT_REF_NAME"
        - git remote set-url origin https://${GL_PROJECT_TOKEN_NAME}:${GL_PROJECT_TOKEN}@${CI_REPOSITORY_URL#*@}
        - semantic-release version
        - |
          if [ "dist" ]; then
            TWINE_PASSWORD=${CI_JOB_TOKEN} TWINE_USERNAME=gitlab-ci-token python -m twine upload --repository-url ${CI_API_V4_URL}/projects/${CI_PROJECT_ID}/packages/pypi dist/*
          fi
      cache:
        paths:
          - ${PIP_CACHE_DIR}
      rules:
        # Don't run on automatic commits
        - if: $CI_COMMIT_AUTHOR =~ /$GL_PROJECT_TOKEN_NAME.*/
          when: never
        # Only run on main/master branch
        - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
  ```

sbates:
I'm very interested in this PR. Does it also imply that a PAT doesn't actually work here? I've been grinding my face on the keyboard all week trying to figure out how to set up auth for Gitlab.

codejedi365:
> Does it also imply that a PAT doesn't actually work here?

No, a Personal, Group or Project Access Token does work if your permissions are correct. 

It is the CI_JOB_TOKEN that does not work inherently due to the pushing of commits and/or tags. CI_JOB_TOKEN's have restricted api endpoint permissions that are not modifiable and unfortunately the ability to write to the remote git repository is not one of them.

Exagone313:
Hello,

Is it possible to re-open this PR, and maybe add the `confirmed` label to prevent GitHub Actions from closing it again?

It's true that it is currently stale as some changes were requested (not having a new provider), but I believe that closing a PR with a good start doesn't help getting progress on the feature.

Thank you :smiley:

codejedi365:
@Exagone313, although I see your point, I don't agree with your solution. My issue is that PRs are restricted to the owner whom opened them and the maintainers. So even if a new contributor came along to complete it they would open a new PR. If they wanted to use the code, they still can from a closed PR by using `git fetch`. It is closed because it's not going to be merged from its state.

In this specific PR, There are current issues that indicate this is still a feature request, so it is not forgotten. I have been monitoring this PR as I still value GitLab usage but I've been working other more significant efforts. Until recently job tokens were not possible as I stated above but luckily changed as identified in #977. I have extracted what code is valuable from this PR and kept it updated on my own branch. But given my feedback previously I can't say this is the code someone else should start with. 

But all in all, a PR will be closed if it is not being worked, because it's a pain to keep it updated as the project improves.

PR #798, opened 2024-01-08
test(fixtures): cache the base example project directory

codejedi365:
## Purpose

## Rationale

In other development towards changelog testing, I came across the high use of IO operations as we rebuild a project with each test case.  This causes a lot of IO operations, and even more so on the use of git.  I followed the rabbit hole, and this PR is the first of a few that help increase the test execution speed through deliberate abstraction/modularity.  I felt it all at once would be way to confusing and too much for a good review.

## How I tested

Besides checking the result here in the CI, I originally was able to confirm the caching mechanism through the VSCode debugger and many break points through the code to see how many times each fixture was called.  Once this commit was implemented, the session scope ensured that it was only called once but the init fixture called every test run.

## How to Verify

To get VSCode to run the debugger, you have to change the `tool.pytest.ini_options.addopts` by removing the `--cov*` options and then changing `-nauto` to `-n0`.  Once you do this, add a breakpoint to the start of `init_example_project`, and `cached_example_project`.  Then execute a test (a minimum of 2 that use `init_example_project`) with the debugger. When you step through, you will see how many times each is called, and you will see the cleanup occur as well if you put a few breakpoints in `teardown_cached_dir`.

### DIFF
--- a/tests/command_line/conftest.py
+++ b/tests/command_line/conftest.py
@@ -53,8 +53,8 @@ def mocked_git_push(monkeypatch: MonkeyPatch) -> MagicMock:
 
 
 @pytest.fixture
-def config_path(example_project: Path) -> Path:
-    return example_project / DEFAULT_CONFIG_FILE
+def config_path(example_project_dir: Path) -> Path:
+    return example_project_dir / DEFAULT_CONFIG_FILE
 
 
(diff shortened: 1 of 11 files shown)


codejedi365:
changed branch name

PR #800, opened 2024-01-09
test(fixtures): refactor to cache base git repo fixture

codejedi365:
## Purpose

## Rationale

PR 2 for test caching. Now the base git repo (git init & configuration on top of the example project) is cached for all the repos to extend from without having to execute the init action every time.  This is phase 2 before the caching of each repo type.

> (from #799) In other development towards changelog testing, I came across the high use of IO operations as we rebuild a project with each test case.  This causes a lot of IO operations, and even more so on the use of git.  I followed the rabbit hole, and this PR is the first of a few that help increase the test execution speed through deliberate abstraction/modularity.  I felt it all at once would be way to confusing and too much for a good review.

This also includes some slight modification to the test fixtures and test setup to better ensure that the files exist as it is not possible to handle when the fixtures resolve/execute when passed as parameters (and their ordering).

## How I tested

Besides checking the result here in the CI, I originally was able to confirm the caching mechanism through the VSCode debugger and many break points through the code to see how many times each fixture was called.  Once this commit was implemented, the session scope ensured that `cached_example_git_project` was only called once but the `git_repo_factory` fixture called every test run by the `repo_*` fixtures.

## How to Verify

To get VSCode to run the debugger, you have to change the `tool.pytest.ini_options.addopts` by removing the `--cov*` options and then changing `-nauto` to `-n0`.  Once you do this, add a breakpoint to the start of `git_repo_factory`, and `cached_example_git_project`.  Then execute a test (a minimum of 2 that use a `repo_*`) with the debugger. When you step through, you will see how many times each is called, and you will see the cleanup occur as well if you put a few breakpoints in `teardown_cached_dir`.

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import os
 import secrets
+import shutil
 import string
 from contextlib import contextmanager
+from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING, Any, Iterable, Tuple, TypeVar
 
@@ -17,6 +19,7 @@ from semantic_release.commit_parser.token import ParseResult
 
 if TYPE_CHECKING:
     import filecmp
+    from pathlib import Path
 
     try:
         from typing import TypeAlias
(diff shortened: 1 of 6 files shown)


codejedi365:
@bernardcooke53, pr ready for review for when you have time, thanks!

tests/util.py
```diff
@@ -28,6 +32,35 @@
...
+        if dest_is_cwd:
+            os.chdir(str(Path(dest).parent))
+
+        shutil.rmtree(dest, ignore_errors=True)
+        shutil.copytree(src=src, dst=dest)
+
+        if dest_is_cwd:
+            os.chdir(dest)
+
+        return
```
bernardcooke53, line 54: Fortunately no longer an issue now 😁

bernardcooke53 (commented):
I like the look of this, need to check it over once more after rebasing but don't see any issues 🚀

tests/util.py
codejedi365: Excellent! Dropped this piece of code!

relekang:
This got merged with #804

PR #804, opened 2024-01-12
test(fixtures): add caching to repo fixtures

codejedi365:
> **INCLUDES** #800, will rebase after merges

**I highly recommend reviewing this PR in the per commit view (click on Commits tab, then commit title of 1st commit, then use next button) as it will make more sense as I moved files around and deliberately made similar changes each commit.**

## Purpose

## Rationale

PR 3 for test caching.  I was able to add caching to all the generated repos on top of the already configured cached projects of example_git_project & example_project.  In doing this, I recognized we have a lot of duplicate code as each commit style variant uses the same git structure and so to simplify the entire repo building process and make it extensible for the future, I used one static definition of all commits mapped to their version and ran it through one git commit engine per structure.

You will also see that I defined the git structures a bit differently to common names used in the industry: Git Flow, GitHub Flow and Trunk Based Development.  In doing so, I refactored the massive `git_repo.py` file into repo submodules categorized by the branch and merge strategy.  This should help provide more extensibility in the future as we already have some examples in issue tickets that aren't addressed. I will be able to address their concerns soon with scenarios.  I do have one more iteration of refactoring to do which will make helper functions for many common git actions but I tried to keep the PR simpler to review by not changing the structure at all (which would cause a bunch of test modifications) but just refactoring the current repo build steps for cross-commit style support.

It may seem a bit confusing at first as now each repo scenario requires 3 fixtures. 1 to build the repo with provided parameters and this fixture has the pre-defined versions and commits per commit style fixtures. The builder fixture (prefixed with `build_`) is to be called from the session level fixture (prefixed `cached_`) which handles building the `cached_files_dir0/cached_$FIXTURE_NAME` base repos.  Then you follow up with the specific test-level fixtures (prefixed `repo_`) which only copy the contents from the cached dir in to the current test's project directory.  I tried to keep the same naming convention of `repo_` to prevent changes across all the tests to simplify the PR.  Some might change in the future.  The result of this structure is that each separate test only copies the files from the cache project it matches to and pytest via `scope="session"` handles creating the cached directory only when requested for the first time. On the second test iteration that calls this repo, no repo is created but just the file path is returned of the cached dir and it copies the contents of the directory over to the new test project directory.  Overall this cuts the test executions speed in half or more depending on the system.  The CI saw an entire minute dropped off, my computer reduced execution time was even more.

> (from #799) In other development towards changelog testing, I came across the high use of IO operations as we rebuild a project with each test case.  This causes a lot of IO operations, and even more so on the use of git.  I followed the rabbit hole, and this PR is the first of a few that help increase the test execution speed through deliberate abstraction/modularity.  I felt it all at once would be way to confusing and too much for a good review.

You do have to modify some fixture helper functions to become `scope="session"` as pytest cannot reference non-session-level scoped fixtures from a session-level scope fixture.  This does mean that a fixture like `update_pyproject_toml` requires the current working directory to be correct as it uses a relative path to the file it needs to modify.

To avoid one of the previous issues with too many open file descriptors due to `git_repo_factory()`, I further considered this concept and instead open and closed the repo object during the session level fixtures.  Then I passed back the Repo object only to the test during the `repo_*` call which ensures it gets closed and only lives open for the duration of the single test.

## How I tested

Besides checking the result here in the CI, I originally was able to confirm the caching mechanism through the VSCode debugger and many break points through the code to see how many times each fixture was called.  Once this PR was implemented, the session scope ensured that `cached_repo_*` functions were only called once but the `repo_*` fixture called every test run.  I further put break points in the builder's and the cached repo functions to ensure the session level fixtures did the caching that was expected.

## How to Verify

To get VSCode to run the debugger, you have to change the `tool.pytest.ini_options.addopts` by removing the `--cov*` options and then changing `-nauto` to `-n0`.  Once you do this, add a breakpoint to the start of `build_*`, `cached_repo_*`, and `repo_*`.  Then execute a test (a minimum of 2 that use a `repo_*`) with the debugger (`test_algorithm_no_zero_dot_versions_angular` would work). When you step through, you will see how many times each is called, and you will see the cleanup occur as well if you put a few breakpoints in `teardown_cached_dir`.


codejedi365:
@bernardcooke53, pr ready for review for when you have time, thanks!

relekang:
Nice 🙌

PR #807, opened 2024-01-15
docs: Remove duplicate note in configuration.rst

tdurk93:
This note is specific to the `commit_author` configuration property, but exists both in its correct place and as a note under the higher-level heading `Settings`


bernardcooke53 (approved):
Thanks @tdurk93!

PR #809, opened 2024-01-17
test(changelog): refactor changelog validation

codejedi365:
> **Recommend** Reviewing this PR through the commit by commit view as opposed to the Files Changed view, it will be easier to understand as you can match my commit message with the changes made. 

## Purpose

1. Enable more granular testing of generated changelog & release notes from the provide templates
2. Fix changelog & release notes to always end with a single newline (not extras)
3. Fix a timezone determination bug which causes failures of changelog tests when executed left of the prime meridian, previous PSR would add time which can impact the date of when a release occurred.

## Rationale

While I've been working on testing some improvements to the changelog functionality, I stumbled across some logic bugs within the testing framework which were producing false positives.  One such example is the regenerate_content from the changelog test. 

1. Refactored some assertions so that the failure is readable and immediately identifiable in the logs rather than obscured.

2. I identified & fixed that some of the angular commit messages that syntactically had the scope in the wrong spot.

3. I ensured that changelogs are actually generated (the empty template directory disabled all default changelog creations) both before the tests and then after the tests.

The strategy I employ here when refactoring the testing for changelogs is to identify what should go as a unit test and what is a scenario based system test.  Personally, I value system tests way more than unit tests. I believe that unit tests really should just focus on "prove it works" with basic default values. The scenarios should be extensive to try all variations as this will catch the most bugs that our users will actually encounter. I have been burned many times when the unit tests don't actually catch the problems users have.  How does this apply to this PR? I removed more parametrize values from unit tests and instead beef up the scenario testing.  One reason I think this applies more here is that the outputs are related to changelog and release notes is that these are the outputs the user receives from the entire program which should be vetted thoroughly as opposed to an internal function.  If you disagree as I realize I am opinionated on this, then I can adjust accordingly.

## How I tested

Related to purpose 3, If you try to execute the test cases on a machine in the Eastern Standard Time (EST) zone after 3PM, all the changelogs tests will fail due to a date mix match. My first assumption was UTC time, however, EST is only 4 or 5 hours off of UTC so at 3PM even UTC isn't on the next day.  But if you interpret a timedelta incorrectly from git as PSR was you instead add 5 hours to UTC time, then you find yourself at 1AM (UTC+5) and Ah Ha!


codejedi365:
@relekang, ready for review

relekang (approved):

PR #847, opened 2024-02-28
Allow special characters in implicit ssh urls

andrew-kline:
Currently, the implicit ssh url regex only allows letters and numbers. These urls can include special characters from email addresses (._-). This corrects that regex and adds an applicable unit test condition.

### DIFF
--- a/semantic_release/helpers.py
+++ b/semantic_release/helpers.py
@@ -114,12 +114,12 @@ def parse_git_url(url: str) -> ParsedGitUrl:
     # Normalizers are a list of tuples of (pattern, replacement)
     normalizers = [
         # normalize implicit ssh urls to explicit ssh://
-        (r"^(\w+@)", r"ssh://\1"),
+        (r"^([\w._-]+@)", r"ssh://\1"),
         # normalize git+ssh:// urls to ssh://
         (r"^git\+ssh://", "ssh://"),
         # normalize an scp like syntax to URL compatible syntax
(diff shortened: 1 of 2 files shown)


semantic_release/helpers.py
```diff
@@ -114,7 +114,7 @@ def parse_git_url(url: str) -> ParsedGitUrl:
     # Normalizers are a list of tuples of (pattern, replacement)
     normalizers = [
         # normalize implicit ssh urls to explicit ssh://
-        (r"^(\w+@)", r"ssh://\1"),
+        (r"^([\w._-]+@)", r"ssh://\1"),
```
codejedi365, line 117: This change to add other special characters to the user name of an implicit ssh URL should also extend to the scp syntax converter below. As after this lines pass, it will now have a `ssh://` in front of it. The point of the next line is to convert the `:` to a `/` so that the ssh url looks like a normal HTTP url for parsing components.

If your unit test worked without this in a bit surprised.

codejedi365 (changes requested):
This PR is on the right track, thank you for the contribution. Just have an additional adjustment to make to the next regex normalizer so that they are congruent.

semantic_release/helpers.py
codejedi365: I made the change which allows the tests to pass.

codejedi365 (approved):

andrew-kline:
@codejedi365, my apologies - your suggestion got lost in my email! Thank you for the change

PR #848, opened 2024-03-04
feat(version-config): add option to disable 0.x.x versions

codejedi365:
## Purpose

- Add a configuration option, `allow_zero_version`, that will toggle the creation of `0.x.x` versions.
- Update documentation to clarify the differences between `allow_zero_version` and `major_on_zero` options

## Rationale

I personally do not publish anything below a `0.x.x` to the world.  With PSR, I have to make an arbitrary breaking change commit that usually isn't a breaking change (It can't break anything if its never been released) to force it to go to `1.0.0`. This PR provides an alternative method where the configuration is set as the minimum release version is `1.0.0`.  

The default of `allow_zero_version` is set to `True` to maintain the current expectations of the project.

The only thing that is a bit different is the situation at the very beginning of a repo where you at least have a `Initial Commit` but not a `feature` or a `fix` but just a chore or non-parsable commit.  At this point, this PR when `allow_zero_version=false`, it will provide back a release of `1.0.0`.  This is purposeful because I am considering that since this is the initial release, its like the initial commit where no real changelog is possible (nothing changes if you aren't released).  When `allow_zero_version=True`, the default version will be `0.0.0` if only chore commits are provided (ie. the original behavior).

## How I Tested

Added an additional test case to target the absolute minimum versions upon repos and updated the zero_dot version testing to include an additional parameter of `allow_zero_version`.  I also added the parameter to `no_zero_dot` versions just for completeness that it does not change any original functionality just like `major_on_zero` is.

> Note that in `test_next_version` there are 4 versions of `test_algorithm_with_zero_dot_versions_*` and `test_algorithm_no_zero_dot_versions_*` with specifics to each of the 4 types of commit messages.  I purely copied and pasted with a find/replace of the commit type, so that should save you some time on the review.

## How to verify

```sh
git fetch origin pull/848/head:pr-848

pytest
```


codejedi365:
@relekang, @bernardcooke53, ready for review

PR #852, opened 2024-03-08
feat(version): Add new version print flags to display the last released version and tag.

cfxegbert:
This adds a the command `semantic-release current-version` so semantic-release can be used to determine the current tag on the branch based on semantic-release config.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -144,6 +144,33 @@ command line options::
     semantic-release version --patch --print
     semantic-release version --prerelease --print
 
+.. _cmd-version-option-print-tag:
+
+``--print-tag``
+***************
+
+Same as the :ref:`cmd-version-option-print` flag but prints the complete tag
+name (ex. ``v1.0.0`` or ``py-v1.0.0``) instead of the raw version number
+(``1.0.0``).
+
+.. _cmd-version-option-print-last-released:
+
+``--print-last-released``
+*************************
+
+Print the last released version based on the Git tags.  This flag is useful if you just
+want to see the released version without determining what the next version will be.
+Note if the version can not be found nothing will be printed.
+
+.. _cmd-version-option-print-last-released-tag:
+
+``--print-last-released-tag``
+***************
+
+Same as the :ref:`cmd-version-option-print-last-released` flag but prints the
+complete tag name (ex. ``v1.0.0`` or ``py-v1.0.0``) instead of the raw version
+number (``1.0.0``).
+
 .. _cmd-version-option-force-level:
 
 ``--major/--minor/--patch``
--- a/semantic_release/cli/commands/version.py
+++ b/semantic_release/cli/commands/version.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, ContextManager, Iterable
 
 import click
 import shellingham  # type: ignore[import]
+from click_option_group import MutuallyExclusiveOptionGroup, optgroup
 from git.exc import GitCommandError
 
 from semantic_release.changelog import ReleaseHistory, environment, recursive_render
@@ -26,8 +27,9 @@ from semantic_release.version import Version, next_version, tags_and_versions
 
 log = logging.getLogger(__name__)
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
     from git import Repo
+    from git.refs.tag import Tag
 
     from semantic_release.cli.config import RuntimeContext
     from semantic_release.version import VersionTranslator
@@ -46,6 +48,13 @@ def is_forced_prerelease(
     """
     log.debug(", ".join(f"{k} = {v}" for k, v in locals().items()))
     return force_prerelease or ((force_level is None) and prerelease)
+
+
+def last_released(
+    repo: Repo, translator: VersionTranslator
+) -> tuple[Tag, Version] | None:
+    ts_and_vs = tags_and_versions(repo.tags, translator)
+    return ts_and_vs[0] if ts_and_vs else None
 
 
 def version_from_forced_level(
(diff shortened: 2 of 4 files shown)


codejedi365:
I'd rather see this be a modification to the version command. Is it possible to extend the current `--print` option funrtion, where your command would be

```sh
semantic-release version --print-current
```

codejedi365:
And also doesn't the --print option do this if you just capture the stdout? It will provide a stderr message but that doesn't stop the return of the current version as long as it hasn't changed?

Or the rationale for this, is to not consider what has changed but only what is currently released?

cfxegbert:
> Or the rationale for this, is to not consider what has changed but only what is currently released?

That is the rationale for this.  I have a need to run my CI/CD pipeline periodically for security checks and I want the current version.

cfxegbert:
Do you think I should be making two version, one that returns the tag and another returning the semantic version?  I could see both being useful.

If I did make a version that printed the tag I would also add a `--print-tag` to be complete.

tests/command_line/test_version.py
```diff
@@ -743,3 +743,35 @@ def test_version_only_update_files_no_git_actions(
...
+    assert result.exit_code == 0
+    assert result.stdout == "0.1.1\n"
+
+
+def test_version_print_last_released_prints_nothing_if_no_tags(
+        repo_with_no_tags_angular_commits: Repo, cli_runner: CliRunner
+):
+    result = cli_runner.invoke(main, [version.name, "--print-last-released"])
+    assert result.exit_code == 0
+    assert result.stdout == ""
```
codejedi365, line 777: These are good tests!  Lets also add a 

```py
assert result.stderr == "No release tags found."
```

docs/commands.rst
```diff
@@ -144,14 +144,30 @@ command line options::
...
 Note if the version can not be found nothing will be printed.
 
+.. _cmd-version-option-print-last-released-tag:
+
+``--print-last-released-tag``
+***************
+
+Same as the :ref:`cmd-version-option-print-last-released` flag but prints the tag instead of the
+semantic version.
+
```
codejedi365, line 170: I would prefer the comment be:

```rst
Same as the :ref:`cmd-version-option-print-last-released` flag but prints the
complete tag name (ex. ``v1.0.0`` or ``py-v1.0.0``) instead of the raw version
number (``1.0.0``).
```

docs/commands.rst
```diff
@@ -144,14 +144,30 @@ command line options::
     semantic-release version --patch --print
     semantic-release version --prerelease --print
 
+.. _cmd-version-option-print-tag:
+
+``--print-tag``
+***************
+
+Same as the :ref:`cmd-version-option-print` flag but prints the tag instead of the semantic version.
```
codejedi365, line 152: same as below just to make sure it is differentiated from what is a tag and what is a version

semantic_release/cli/commands/version.py
```diff
@@ -117,11 +120,25 @@ def shell(cmd: str, *, check: bool = True) -> subprocess.CompletedProcess:
         "help_option_names": ["-h", "--help"],
     },
 )
-@click.option(
+@optgroup.group("Print flags", cls=MutuallyExclusiveOptionGroup)
```
codejedi365, line 123: Very neat!

codejedi365 (commented):
Thanks for the contribution. Just a few recommendations but I'm glad you were thorough with tests and doc updates.

One minor recommendation. I would rewrite your branch commits to remove at least the first one and then separate out the addition of the `click-option-group` as its own `build(deps): add click-option-group ...` because the maintainer tends to rebase and merge with the exact commits used. Once this occurs then a changelog is generated from those commits. If your first commit is in there then that is confusing to what actual features were added and the build change is then hidden from the changelog as its lumped into the feature creation.

codejedi365:
> Do you think I should be making two versions, one that returns the tag and another returning the semantic version? I could see both being useful.
> 
> If I did make a version that printed the tag I would also add a `--print-tag` to be complete.

Adding all 4 variants was a great choice, thanks for thinking ahead.

semantic_release/cli/commands/version.py
```diff
@@ -48,6 +49,11 @@ def is_forced_prerelease(
     return force_prerelease or ((force_level is None) and prerelease)
 
 
+def last_released(repo: Repo, translator: VersionTranslator) -> Optional[tuple[Tag, Version]]:  # noqa: UP007
```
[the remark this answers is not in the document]
cfxegbert: This is supposed to support python3.8 which does not have the '|' syntax.  I think it was added in 3.10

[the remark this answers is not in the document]
codejedi365: > This is supposed to support python3.8 which does not have the '|' syntax.  I think it was added in 3.10

Good point, we do support 3.8 still. However the line 1 import of __future__.annotations seems to resolve all of that. Through the rest of the code it hasn't been a problem

cfxegbert:
Addressed changes and cleaned up commit log

codejedi365:
@cfxegbert, good job thanks!

This is an optional change as it doesn't change much for your overall PR, but I wanted to recommend that the `pre-commit` commit should be of type `chore` instead. As this software is all about version detection, the `fix(pre-commit)` commit by itself would actually trigger a patch release of the codebase but `chore` would not. You would want to reserve `feat` and `fix` for specific commits that are related to the source code that is published/consumed by end users. A development environment change does not affect end user so you would not want to cause a patch release itself. This is less relevant for this specific PR as you have an independent feature commit as well so that overrides the fix level commit previously made. 

All in all thanks for the effort you have put into this PR!

@relekang, PR ready for merge.

cfxegbert:
Fix commit messages not to start with a capitol letter and not end with a period

codejedi365 (approved):
Great work! Approved.

Ready for merge @relekang

relekang (approved):

PR #860, opened 2024-03-15
test(changelog): use UTC times in unit tests

cfxegbert:
The unit tests in changelog where using the local timezone.  The change log and release notes are in UTC

### DIFF
--- a/tests/unit/semantic_release/changelog/test_default_changelog.py
+++ b/tests/unit/semantic_release/changelog/test_default_changelog.py
@@ -1,5 +1,5 @@
 # NOTE: use backport with newer API
-from datetime import datetime
+from datetime import datetime, timezone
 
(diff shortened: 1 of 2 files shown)


codejedi365:
Thank you for your time in trying to fix this issue. I actually thought this was the solution too a while back but ends up its actually a timestamp interpretation problem inside of semantic-release.  We just don't have good testing of changelog generation which is why the UTC fix seems to work for you.  I previously handled this in the PR #809 that has been awaiting merge since 1 March.  I recommend building off of that PR branch if you are doing any changelog validation stuff or at least cherry-pick [d6e15fe](https://github.com/python-semantic-release/python-semantic-release/pull/809/commits/d6e15fe41a40e244e216f44f73302ae2284d60c0).

PR #867, opened 2024-03-23
fix: issue vs PR/MR urls

codejedi365:
## Purpose

Not all HVCS call Pull Requests the same way as GitHub.  GitHub considers an PR another issue and allows for the same url resolution of PRs as issues.  GitLab on the other hand uses a separate number scheme and url to match Merge Requests as separate than issues.

## Rationale

I corrected the implementation of GitLab to have separate MR and issue url generation.  This meant that a new filter needed to be past to the changelog context and all the other hvcs needed to inherit/override the same issue function.  The unit tests were updated to generate these urls properly across all HVCS's.  

## How I tested

- Must add a test that actually adds a commit that hvcs should apply a url to however our testing does not currently provide this.

### DIFF
--- a/semantic_release/changelog/context.py
+++ b/semantic_release/changelog/context.py
@@ -31,5 +31,9 @@ def make_changelog_context(
         repo_name=hvcs_client.repo_name,
         repo_owner=hvcs_client.owner,
         history=release_history,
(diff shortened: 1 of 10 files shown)


codejedi365:
wrapped into #890

Issue #906, opened 2024-05-01
Support tag in GH action config


rantoniuk:
`--commit` and `--tag` are supported separately, i.e. I can run a release on a non-PyPi project that does not require any source code changes and just create a tag and pushing it when doing a release.

This is possible in the tool:
https://python-semantic-release.readthedocs.io/en/latest/commands.html#tag-no-tag

but it's not possible in GH action, as there is only `commit` option available:
https://github.com/python-semantic-release/python-semantic-release/blob/master/action.yml#L53
and no `tag` option.

codejedi365:
That is a good catch, thank you. I did need to up date it.  It is not easy keeping up with the action config as there isn't real testing for it.

codejedi365:
@rantoniuk , would you try out this fix and see if it works for you?

```yml
jobs:
  release:
    runs-on: ubuntu-latest
    concurrency: release
    permissions:
      id-token: write
      contents: write

    steps:
    - uses: actions/checkout@v3
      with:
        fetch-depth: 0

    - name: Python Semantic Release
      uses: codejedi365/python-semantic-release@fix/gha-missing-tag-option
      with:
        github_token: ${{ secrets.GITHUB_TOKEN }}
        commit: false
        tag: true
        # ...
```

rantoniuk:
Thanks for quick reply. Actually I tested it using the current version and it works without that patch as expected, i.e. it creates a tag but does not push any commits.
I left a PR comment as well :)

codejedi365:
> I left a PR comment as well :) 

Thank you!

> Actually I tested it using the current version and it works without that patch as expected, i.e. it creates a tag but does not push any commits.

I realize now that really it's because the default is `--commit` and `--tag` and if you just turn off commit with `commit: false` then it is implicitly `--tag` but there is no explicit option to toggle the tag flag via the GitHub Action until my PR. So the PR gives the explicit ability to toggle the tag and commit options separately.

codejedi365:
> Quick question, how do you auto-comment the related github issues to add the release information (above)?

Manual copy and paste! lol! 

I copied the template from the JS version of semantic-release, which  auto-posts messages like these for all issues that were resolved, because of how awesome it is. One day, PSR will too.

PR #908, opened 2024-05-02
fix(gha): add missing `tag` option to GitHub Action definition

codejedi365:
## Purpose

- Update the GitHub action definitions to include the `--tag` command line option added in `v8.4.0`
- Resolve: #906

## Rationale

First I must add the definition to the `action.yml` file to ensure that users can provide a `tag` directive to their `GitHub Action` definition.  This then translates to a new environment variable that is passed to the `action.sh`.  That variable must be read in order to determine if it should add a true or false command line option to the `semantic-release version` command.  Lastly, I decided to remove all the extra default values from the `action.*` files because they we are maintaining multiple locations where the defaults are set for the project.

### DIFF
--- a/action.sh
+++ b/action.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-# Convert "true"/"false" into command line args
+# Convert "true"/"false" into command line args, returns "" if not defined
 eval_boolean_action_input() {
 	local -r input_name="$1"
 	shift
@@ -12,7 +12,9 @@ eval_boolean_action_input() {
 	shift
 	local -r if_false="$1"
 
-	if [ "$flag_value" = "true" ]; then
+	if [ -z "$flag_value" ]; then
+		echo ""
+	elif [ "$flag_value" = "true" ]; then
 		echo "$if_true"
 	elif [ "$flag_value" = "false" ]; then
 		echo "$if_false"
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -56,15 +56,21 @@ inputs:
     default: "true"
     description: Whether or not to commit changes locally
 
+  tag:
+    type: string
+    required: false
+    default: "true"
```
rantoniuk, line 62: This should be true by default:
https://python-semantic-release.readthedocs.io/en/latest/commands.html#tag-no-tag

action.sh
```diff
@@ -34,6 +34,7 @@ export GIT_COMMIT_AUTHOR="${GIT_COMMITTER_NAME} <${GIT_COMMITTER_EMAIL}>"
 export ROOT_OPTIONS="${INPUT_ROOT_OPTIONS:="-v"}"
 export PRERELEASE="${INPUT_PRERELEASE:="false"}"
 export COMMIT="${INPUT_COMMIT:="false"}"
+export TAG="${INPUT_TAG:="false"}"
```
rantoniuk, line 37: This should be true by default:
https://python-semantic-release.readthedocs.io/en/latest/commands.html#tag-no-tag

And the line above as well is incorrect per action.yml spec.

codejedi365: I think I'm going to just remove all bash expansion defaults as defaults are handled within PSR itself rather than trying to maintain these environment variables from GitHub Actions in line with the defaults of the project. Well then we also put defaults in the `action.yml` too.

PR #921, opened 2024-05-08
chore: bump build version in pyproject.toml and contributor guide

afuetterer:
Hi there,

I was unsure, why the build version is pinned to such an old version. Is there a reason?

What do you think about also adding `build[uv]` and using `python -m build --installer=uv .`?

Ref: https://build.pypa.io/en/stable/#python--m-build---installer

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -33,13 +33,13 @@ Install this module and the development dependencies
 
 .. code-block:: bash
 
-    pip install -e ".[dev,mypy,test]"
+    pip install -e .[dev,mypy,test]
 
 And if you'd like to build the documentation locally
 
 .. code-block:: bash
 
-    pip install -e ".[docs]"
+    pip install -e .[docs]
     sphinx-autobuild --open-browser docs docs/_build/html
 
 Testing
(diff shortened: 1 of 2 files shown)


codejedi365:
Hi @afuetterer,

> I was unsure, why the build version is pinned to such an old version. Is there a reason?

Not that I'm aware of. I think it is a by-product of not tracking the updates of the dependency especially when it is defined outside of our normal dependencies list.  We have dependabot configured on the repository to automatically open PRs for bumps of dependencies but it wouldn't know to look in the `build_command` to detect a need to open a PR. I'm glad you brought it up as it would have gone unnoticed for quite some time.  

In lieu of what you suggested, I added on top of your branch to move the build dependency up into an optional dependency definition to prevent the problem for the future. Now Dependabot will open PRs as it does with the rest of the dependencies.  Obviously this also triggered slight adjustment to the build commands to point at the new optional dependency key which improves overall maintainability.

> What do you think about also adding `build[uv]` and using `python -m build --installer=uv .`?

As for `uv` usage, this was new to me and I'm glad to find out about a highly performant & improved `pip`, however, it is still too nascent for me to trust to not cause other issues on this very popular library.  I would rather wait until it releases a `1.0.0` at least.

afuetterer:
Thanks. You are right, the change I suggested would get outdated over time again. Good point to have build be monitored by dependabot as well.

PR #929, opened 2024-05-15
add prelease-token in github action.sh

glichtner:
# Problem
The github action doesn't translate the prelease_token option from the action.yml into the corresponding command line parameter. 

# Fix 
Add the parameter to ARGS in action.sh.


codejedi365 (approved):

codejedi365:
@glichtner, thanks!

Issue #721, opened 2023-09-26
ValueError: SHA b'f697dc989970b..9' could not be resolved, git returned: b'f697dc989970b..9 missing'


VathsalaM:
Hi,

I am trying to use this package in github actions. But, I get the above error. I verified that the commit actually exists. May I know what am I missing here? 
<img width="366" alt="Screenshot 2023-09-26 at 2 39 38 PM" src="https://github.com/python-semantic-release/python-semantic-release/assets/15318377/599696fd-d44e-4b5d-b76f-3b18cbba551b">

```name: Github Packages Version Increment Actions
on:
  push:
    branches: [ main ]

jobs:
  Build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-python@v4
        with:
          python-version: 3.9
      - name: Install Python Poetry
        uses: abatilo/actions-poetry@v2.3.0
        with:
          poetry-version: 1.6.1
      - name: Configure poetry
        shell: bash
        run: python -m poetry config virtualenvs.in-project true
      - name: View poetry version
        run: poetry --version
      - name: Install dependencies
        run: |
          python -m poetry install

  Release:
    needs: Build
    # https://github.community/t/how-do-i-specify-job-dependency-running-in-another-workflow/16482
    if: github.event_name == 'push' && github.ref == 'refs/heads/main' && !contains(github.event.head_commit.message, 'chore(release):')
    runs-on: ubuntu-latest
    permissions:
      packages: write
      contents: write
      issues: write
      pull-requests: write
    steps:
      - uses: actions/setup-python@v4
        with:
          python-version: 3.9
      - name: Checkout code
        uses: actions/checkout@v3
      - name: Semantic Release
        run: |
          pip install python-semantic-release
          git config user.name github-actions
          git config user.email github-actions@github.com
          semantic-release version --print
          ```

kamoo1:
Hi

I just fixed this exact error by adding `fetch-depth: 0` to the checkout action.

gurland:
I am getting the same error using Gitlab CI. python-semantic-release just tries to resolve unknown random SHA

PR #930, opened 2024-05-15
chore: attempt to help the user if failure finding commits

JohnVillalovos:
I experienced a failure due to not setting the fetch-depth to 0 (everything).

It was like this:
  ValueError: SHA b'1234567890123456789012345678901234567890' could not be resolved, git returned: b'1234567890123456789012345678901234567890 missing'

Now print an error suggesting to the user they check their git fetch depth settings.

Related: #721

### DIFF
--- a/semantic_release/version/algorithm.py
+++ b/semantic_release/version/algorithm.py
@@ -113,13 +113,21 @@ def _bfs_for_latest_version_in_history(
             log.debug("commit %s doesn't match any tags", node.hexsha)
 
             # Add all parent commits to the queue if they haven't been visited
-            for parent in node.parents:
-                if parent in visited:
-                    log.debug("parent commit %s already visited", node.hexsha)
-                    continue
+            try:
+                for parent in node.parents:
+                    if parent in visited:
+                        log.debug("parent commit %s already visited", node.hexsha)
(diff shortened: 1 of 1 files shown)


codejedi365:
@JohnVillalovos, thank you for time you put into solving this issue for folks and I'm sorry for the delay in response.  

In relation to your recommended fix, I would rather solve the primary issue than have a tailored error message especially since we support more than just GitHub and GitLab CI environments.  The fix I would like to have is prior to the algorithm search is to check with `git` for is it a `shallow` repository.  If it is then run `git fetch --unshallow` to retrieve the full repository.  This behavior is more active to ensure the proper environment in which we require to successfully determine the version and generate the changelog.  Any time a git clone is created with a limited depth, git defines the repository as shallow and can be queried.

Let me know if you would be willing to implement this adjustment.

JohnVillalovos:
@codejedi365 Thanks for the reply. At the moment I don't think I have the motivation/time to dig into the code-base to figure how to do that.

So I will let someone else take up that task.

Thanks again.

codejedi365:
Not a problem, thanks for the honesty. I'll find time to add it in over the next few months.

PR #942, opened 2024-05-30
fix: fix problems with windows commands

JCHacking:
Depending on the command it may fail in windows because:

- The SYSTEMROOT environment variable (refers to the windows installation path C:Windows) is missing.
- The WINDIR environment variable (Same as SYSTEMROOT) is missing.

I think it makes sense to include it by default instead of adding it when they have problems because it is not an environment variable as it could be SLOW_TEST=True or something like that.


codejedi365:
@JCHacking please update the test `test_version_runs_build_command` with your windows env var. We only want to validate that a set env var is passed to the build command. You do not need to add this to the following test of user defined env vars.

Please also document your newly passed env variable to the `build_command` table of all env variable definitions on `docs/configuration.rst`. Please include a windows relevant clause and what would happen on non-windows systems.

JCHacking:
Hi @codejedi365 
Added in the tests and in the documentation.
I will check if there are any other windows environment variables that deserve to be added.

In the case of non-windows systems, it is normal that none of these systems are present, so they will not be imported.
Would it be worth adding a validation of whether the platform is windows to add these? Or do we understand that the probability is very small?

JCHacking:
I have been looking at the poetry and pdm code to see what environment variables they use and get an idea. There is a lot of variety.

These are the variables documented by microsoft https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables

Do I add them all?

codejedi365:
> In the case of non-windows systems, it is normal that none of these systems are present, so they will not be imported.

Agreed; I was just recommending this be included in the docs.

> Would it be worth adding a validation of whether the platform is windows to add these? Or do we understand that the probability is very small?

No, I don't think so. The current code uses a None Filter so if they don't exist they won't be passed through anyway. Even if they do exist it won't cause much difference. The reason I am selectively passing them is to ensure no secrets are accidentally passed on that are not explicitly set by the user in `build_command_env`.

codejedi365:
> Do I add them all?

Great find, and no, that sounds like overkill. Let's break out a function for windows env as I mentioned in my code comment, and then only provide the non-CSIDL prefixed vars.

JCHacking:
- Extracted to another method 
- Added the rest of environment variables
- Add only on windows systems

It is not documented but I have added PATHEXT, this environment variable tells which are the extensions of the files that can be executed, I think it is useful.

If you are ok with the way the code is now, I think about how we could test this (probably parameterized with sys.platform).

JCHacking:
Hi @codejedi365 
I have added the tests and documentation.

The test `test_version_runs_build_command` now says it is linux and the system has windows environment variables to make sure that in case it is a non-windows system they are not passed.

And I have created another test `test_version_runs_build_command_windows` with the scenario of a windows system.

codejedi365:
> It is not documented but I have added PATHEXT, this environment variable tells which are the extensions of the files that can be executed, I think it is useful.

It seems to be documented here related to powershell: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.4

tests/command_line/test_version.py
```diff
@@ -652,6 +676,107 @@ def test_version_runs_build_command(
...
+                "SYSTEM": "C:\\Windows\\System32",
+                "SYSTEM16": "C:\\Windows\\System16",
+                "SYSTEM32": "C:\\Windows\\System32",
+                "SYSTEMDRIVE": "C:",
+                "SYSTEMROOT": "C:\\Windows",
+                "TEMP": "C:\\Users\\Username\\AppData\\Local\\Temp",
+                "TMP": "C:\\Users\\Username\\AppData\\Local\\Temp",
+                "USERPROFILE": "C:\\Users\\Username",
+                "USERSID": "S-1-5-21-1234567890-123456789-123456789-1234",
+                "WINDIR": "C:\\Windows"
```
codejedi365, line 776: If you could have all these reference the `patched_os_environment["VAR_NAME"]` so that it is easier to maintain and it is congruent with the rest of it, that would be great!

codejedi365 (approved):
Great work, thank you for putting in the time and effort.  Just a few aesthetics remain, sorry for the delay in reply.

JCHacking:
Formatted with ruff and changed tests to use patch_os_environment.
Are there any other changes needed? I think I read about using the windows check in get_windows_env, but now I can't find the message, do you want me to change it?

codejedi365 (approved):
Great thank you for the updates.  We can leave what you have for the `is_windows` ternary statement.

JCHacking:
Great, then I see it as ready for me too.

Let me know when this is versioned so I can update the library in my developments.

Thank you very much!!!

Issue #952, opened 2024-06-15
Default changelog template options


wyardley:
## Description
The default changelog template should provide an easy way for users to get things like the short description (subject of the commit vs. all lines of the commit message), the type, scope, etc. on their own.

This would avoid unexpected behavior like what's seen in the existing `python-semantic-release` changelog, with things like `Co-Authored-By:` showing up in the generated markdown, and would make the changelog more readable in cases where there are bulleted lists in the body of the commit, since the changelog itself is also bulleted. It would be nice to be able to get the scope, type, and short message via variables vs. having to do all of this within the template (see #924 for example)

It should also force a specific style of newline (currently, if the commits have `^M`s in them, it seems like they'll end up in the generated changelog).

## Use cases
https://github.com/python-semantic-release/python-semantic-release/blob/master/CHANGELOG.md

Avoid confusing stuff like this in the changelog (where the `------` gets converted into something that breaks up the markdown, and where things like `Co-Authored-By` get inserted in randomly).
<img width="1118" alt="image" src="https://github.com/python-semantic-release/python-semantic-release/assets/10714426/3942a794-88d5-4cf1-af92-ee34527983e8">

Also, you could potentially do something like this
```
* {%-if foo.scope -%}**{{ foo.scope }}**: {%- endif -%}{{ foo.short_message }} ([`{{ commit.commit.hexsha[:7] }}`]({{ commit.commit.hexsha | commit_hash_url }}))
```
(not sure if that would be within `commit` or some other hierarchy) to get
```
* **foobar**: add xyz ([`93e4cca`](https://github.com/someorg/somerepo/commit/93e4ccab200eaca9e0fbaa0fd52be0656b6165c0))
```
from the input
```
feat(foobar): add xyz

* Here is a longer set of bulleted or non bulleted descriptions from the body of
   the commit
```

(note that type is stripped and scope is now available separately)

wyardley:
(side note: using the builtin templating, something like this works, though would probably be cleaner to make a macro / function vs. repeating it)
`commit.commit.message.rstrip().partition('\n')[0] }}`

wyardley:
Something like this basically does what I want:

```python3
{% for commit in commits %}{% if type_ != "unknown" %}
* {% if commit.scope %}**{{ commit.scope }}**: {% endif %}{{ commit.descriptions[0] }} ([`{{ commit.short_hash }}`]({{ commit.hexsha | commit_hash_url }}))
{% else %}
[...]

PR #956, opened 2024-06-15
fix(commit_parser): strip DOS carriage-returns in commits

### LINKED ISSUE #955: Inconsistent carriage returns in generated changelogs
## The problem
The commit object returned from GitPython seems to sometimes contain DOS CRs (`^M`)s (even when the original commit message does not contain them). This results in inconsistent carriage return style in generated changelogs with the default template

## Expected behavior
Consistent CRs in generated changelog

## Environment
```shell
% python3 --version
Python 3.12.3
% pip3 --version   
pip 24.0 from /opt/homebrew/lib/python3.12/site-packages/pip (python 3.12)
% semantic-release --version
semantic-release, version 9.8.1
```

I can make a PR to fix this, though may need some help with figuring out the best way(s) to test if applicable.

wyardley:
The default template can result in mixed (UNIX / DOS style) carriage returns in the generated changelog. Use a string replace in the commit parser to strip the DOS CRs ("\r"). While it's possible to do the replacement in the template, IMO, this is the better / correct way to fix it.

I checked, and seems like maybe it's only needed when we're _not_ doing a byte decode (so the case where `m` is a raw string)? I can add it back to both if you'd like, though.

Fixes #955
Note: this may result in some diffs in generated changelogs.

See also #952 

I'm happy to adjust existing tests in a way that would catch this and / or add a new test if someone can give me a general pointer on the best way / best place to do this. Also feel free to suggest a terser or more idiomatic way to do this.

### DIFF
--- a/semantic_release/commit_parser/token.py
+++ b/semantic_release/commit_parser/token.py
@@ -21,7 +21,8 @@ class ParsedCommit(NamedTuple):
     @property
     def message(self) -> str:
         m = self.commit.message
-        return m.decode("utf-8") if isinstance(m, bytes) else m
+        message_str = m.decode("utf-8") if isinstance(m, bytes) else m
+        return message_str.replace("\r", "")
 
     @property
     def hexsha(self) -> str:
@@ -39,7 +40,8 @@ class ParseError(NamedTuple):
     @property
     def message(self) -> str:
         m = self.commit.message
-        return m.decode("utf-8") if isinstance(m, bytes) else m
+        message_str = m.decode("utf-8") if isinstance(m, bytes) else m
+        return message_str.replace("\r", "")
 
     @property
     def hexsha(self) -> str:


semantic_release/commit_parser/token.py
```diff
@@ -21,7 +21,7 @@ class ParsedCommit(NamedTuple):
     @property
     def message(self) -> str:
         m = self.commit.message
-        return m.decode("utf-8") if isinstance(m, bytes) else m
+        return m.decode("utf-8") if isinstance(m, bytes) else m.replace("\r", "")
```
wyardley, line 24:
```suggestion
        message_str =  m.decode("utf-8") if isinstance(m, bytes) else m
        return message_str.replace("\r", "")
```

Could also do it this way

codejedi365:
I know this seems a bit like a crazy idea but maybe we don't strip the `\r` out but instead format the changelog with `\r\n` or windows style, because then it is always readable cross-platform.  If we remove the `\r` then we restrict it only to Linux machines for reviewing the changelog and it does not work inheritly on windows. However the reverse will work on both.  

Maybe this actually just bleeds into the code that handles the rendering rather than the content of the commit message. So it is likely your modification to remove them essentially is correct, then the during the writing of the file is when we apply the `\r\n`'s.

wyardley:
I don’t think that’s necessary? The rest of the codebase _and_ most of the generated changelog is already using regular UNIX style CRs and it doesn’t sound like that’s caused any issues for people?

codejedi365:
I noticed that if someone uses the GitHub editor to make a commit that the commit message from the text area in the webUI will create commits with `/r/n`. That is more likely where those odd ones are coming from in our changelog.

semantic_release/commit_parser/token.py
codejedi365: This is probably the better way to catch it regardless if it is bytes or not.

codejedi365 (changes requested):
I like the second way as stated above. Thanks for the effort

semantic_release/commit_parser/token.py
wyardley: Yeah... in my tests, it only seemed to be necessary for the string case, but agree with you this is probably better to be consistent and to be sure.

wyardley:
> I like the second way as stated above. Thanks for the effort

Updated.

> I noticed that if someone uses the GitHub editor to make a commit that the commit message from the text area in the webUI will create commits with /r/n.

Now _this_ makes a lot of sense, and that's probably the reason.

codejedi365 (approved):

codejedi365 (approved):

PR #957, opened 2024-06-15
chore: update pre-commit hooks

wyardley:
Note: the current `ruff` pre-commit checks check `.`, however, there are multiple existing errors. Maybe it would make sense to switch `ruff` to run on changed files only unless `--all-files` is passed? That would be gentler, especially if we're not requiring all these checks to pass (that said, #958 has some misc fixes that resolve _some_ of the outstanding flagged and non-ignored issues).
 
- Update pre-commit hooks
- Switch away from deprecated flag to `ruff`
- Resolve two minor pre-commit hook findings

Note: I still see the following finding from `mypy`:
```
semantic_release/cli/commands/publish.py: note: In function "publish":
semantic_release/cli/commands/publish.py:66:9: error: "HvcsBase" has no
attribute "upload_dists"  [attr-defined]
            hvcs_client.upload_dists(tag=tag, dist_glob=pattern)
            ^~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 49 source files)
```
Maybe an inheritance issue with it being in `remote_hvcs_base.py` vs `_base.py`?

There is also some commented out code that ruff is flagging - I can remove in another PR if you'd like.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,15 +13,13 @@ repos:
 
   # Security & credential scanning/alerting
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v4.6.0
     hooks:
       - id: debug-statements
       - id: detect-aws-credentials
         args: ["--allow-missing-credentials"]
       - id: detect-private-key
       - id: check-builtin-literals
-      - id: check-executables-have-shebangs
-      - id: check-shebang-scripts-are-executable
       - id: check-yaml
       - id: check-toml
       - id: check-case-conflict
@@ -31,16 +29,6 @@ repos:
       - id: mixed-line-ending
       - id: check-ast
 
-  # Formatters that may modify source files automatically
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Keep old ruff formatting rules until other merge requests are completed
-    rev: v0.1.11
-    hooks:
-      - id: ruff-format
-        name: ruff (format)
-        args: ["."]
-        pass_filenames: false
-
   - repo: https://github.com/adamchainz/blacken-docs
     rev: 1.16.0
     hooks:
@@ -48,14 +36,14 @@ repos:
         additional_dependencies: ["black==23.10.1"]
 
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.15.1
+    rev: v3.16.0
     hooks:
       - id: pyupgrade
         args: ["--py38-plus", "--keep-runtime-typing"]
 
   # Linters and validation
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.2
+    rev: v0.4.9
     hooks:
       - id: ruff
         name: ruff (lint)
@@ -63,15 +51,14 @@ repos:
           - "--fix"
           - "--exit-non-zero-on-fix"
           - "--statistics"
-          - "--output-format=text"
-          - "."
-        pass_filenames: false
+          - "--output-format=full"
+      - id: ruff-format
+        name: ruff (format)
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: "v1.9.0"
+    rev: "v1.10.0"
     hooks:
       - id: mypy
-        args: ["--config-file", "pyproject.toml"]
         additional_dependencies:
           - "pydantic>=2,<3"
           - "types-requests"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -282,7 +282,6 @@ ignore = [
 ]
 
 external = ["V"]
-ignore-init-module-imports = true
 task-tags = ["NOTE", "TODO", "FIXME", "XXX"]
 
 [tool.ruff.format]
--- a/semantic_release/cli/config.py
+++ b/semantic_release/cli/config.py
@@ -558,7 +558,7 @@ class RuntimeContext:
             # with our forced empty value at the end which can be dropped
             parts = [*env_var_def.split("=", 1), ""]
             # removes any odd spacing around =, and extracts name=value
-            name, env_val = [part.strip() for part in parts[:2]]
+            name, env_val = (part.strip() for part in parts[:2])
 
             if not name:
                 # Skip when invalid format (ex. starting with = and no name)
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -73,7 +73,6 @@ jobs:
       - name: ruff
         run: |
           python -m ruff check . \
-            --config pyproject.toml \
             --diff \
             --output-format=full \
             --exit-non-zero-on-fix
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -107,7 +107,7 @@ If a setting is not provided, than PSR will fill in the value with the default v
 Python Semantic Release expects a root level key to start the configuration definition. Make
 sure to use the correct root key dependending on the configuration format you are using.
 
-.. note:: If you are using ``pyproject.toml``, this heading should include the `tool`` prefix
+.. note:: If you are using ``pyproject.toml``, this heading should include the ``tool`` prefix
           as specified within PEP 517, resulting in ``[tool.semantic_release]``.
 
 .. note:: If you are using a ``releaserc.toml``, use ``[semantic_release]`` as the root key
--- a/semantic_release/cli/commands/publish.py
+++ b/semantic_release/cli/commands/publish.py
@@ -63,4 +63,4 @@ def publish(cli_ctx: CliContextObj, tag: str = "latest") -> None:
 
     log.info("Uploading distributions to release")
     for pattern in dist_glob_patterns:
-        hvcs_client.upload_dists(tag=tag, dist_glob=pattern)
+        hvcs_client.upload_dists(tag=tag, dist_glob=pattern)  # type: ignore[attr-defined]


semantic_release/cli/config.py
```diff
@@ -558,7 +558,7 @@ def from_raw_config(
             # with our forced empty value at the end which can be dropped
             parts = [*env_var_def.split("=", 1), ""]
             # removes any odd spacing around =, and extracts name=value
-            name, env_val = [part.strip() for part in parts[:2]]
+            name, env_val = (part.strip() for part in parts[:2])
```
wyardley, line 561: This was added by `pyupgrade`

.pre-commit-config.yaml
```diff
@@ -48,27 +48,27 @@ repos:
...
+    rev: v0.4.9
     hooks:
       - id: ruff
         name: ruff (lint)
         args:
           - "--fix"
           - "--exit-non-zero-on-fix"
           - "--statistics"
-          - "--output-format=text"
+          - "--output-format=full"
```
wyardley, line 66: updated elsewhere in b41ad31793e5ee02359e8949cc0dd5d69240b622 I think.

.pre-commit-config.yaml
```diff
@@ -31,44 +31,38 @@ repos:
...
 
   # Linters and validation
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.2
+    rev: v0.4.9
     hooks:
+      - id: ruff-format
+        name: ruff (format)
+        args: ["."]
+        pass_filenames: false
```
wyardley, line 53: I think removing these two lines for both lint and fmt should switch it back to the more traditional behavior of only running on modified files, which IMO is the right behavior for a pre-commit hook (esp. since the checks run on everything in CI anyway)

action.sh
codejedi365, line 1: Looking at the chmod value, please do not add execute to this. It is not intended to be run in someone's environment and I don't like laying down executable shell scripts on someones box via git.  Making it executable is not needed here because the Dockerfile calls `bash action.sh` which means it doesnt require execution permissions but still runs as intended.

.pre-commit-config.yaml
codejedi365: yes that is correct

semantic_release/cli/config.py
codejedi365: makes sense

action.sh
wyardley: Ok - I was basing this on the existing pre-commit hooks existing. So I think if you don't want this behavior, the repo should not have those lines in the pre-commit config either. I removed them and made the script non-executable.

.pre-commit-config.yaml
codejedi365: concur. only lets keep it the same version that is in `pyproject.toml` for now, and then also see if there is a dependabot configuration for this file.  That is how we keep ruff updated in `pyproject.toml`. There is a PR for `ruff` currently from dependabot but i haven't merged it yet because honestly `ruff` is releasing so frequently it's annoying.  I'm slow rolling the updates here for this dependency.  Just for awareness, I have it locked to the specific patch version because if there is a bug fix that re-enables a rule it would arbitrarily cause the CI to break upon their release.  I'm not okay with that so I'll just slow roll the merges as its only a dev dependency.

action.sh
codejedi365: Good catch, yeah I haven't migrated over to use `pre-commit` since becoming maintainer of this project so I was unaware where that was coming from.

.pre-commit-config.yaml
wyardley: I don't think the hook version corresponds directly to the ruff version, though they may still be linked.

Not 100% sure on dependabot, but FWIW, Renovate does have experimental support (which works quite well) for keeping track of pre-commit hook versions.

I saw the comment, but I actually think (could be wrong) that having the old version of the pre-commit pinned here might be part of why people are seeing some fighting sometimes (as maybe the hook uses an internal version of `ruff`) tied to its own version?

I do see errors with the current ruff version if I run `ruff check` locally as well, though (note: using Python 3.12 vs. 3.8, but the toml config does have target version set to 3.8) - do you not see that?

wyardley: per https://github.com/python-semantic-release/python-semantic-release/pull/958#issuecomment-2170391124, adjusted to operate only on changed files. will double-check the behavior.

pyproject.toml
```diff
@@ -282,7 +282,6 @@ ignore = [
 ]
 
 external = ["V"]
-ignore-init-module-imports = true
```
wyardley, line 285: this resolves this deprecation warning:
```
warning: The `ignore-init-module-imports` option is deprecated and will be removed in a future release. Ruff's handling of imports in `__init__.py` files has been improved (in preview) and unused imports will always be flagged.
```

wyardley:
However, the hook version doesn't seem to affect the version of `ruff` run. _if_ I use `--all-files`, I do still see a couple errors with `ruff-lint`, as well as one `mypy` error that doesn't seem to show up in CI. Either way, I think this will solve at least some of the problems with running it locally, but let me know if you want to pin back the pre-commit hook to match the ruff version.

codejedi365 (approved):
I bumped the project ruff version so that matches what is in the pre-commit definition. And then rebased this branch so that they are the same.

Also I did review the `ruff-pre-commit` releases to find that each version corresponds to which ruff it uses under the hood.

codejedi365:
> However, the hook version doesn't seem to affect the version of `ruff` run. _if_ I use `--all-files`, I do still see a couple errors with `ruff-lint`, as well as one `mypy` error that doesn't seem to show up in CI. Either way, I think this will solve at least some of the problems with running it locally, but let me know if you want to pin back the pre-commit hook to match the ruff version.

I really believe this is because the pre-commit is not reading our config in the pyproject.toml file because if you see the lint job in the ci we have to specify the configuration file to ruff. Why it doesn't do this by default is besides me but it changes what is flagged in the CI

wyardley:
Also, just read that the lint should come before format if lint is run with `—fix`, so maybe I should switch the order?

wyardley:
Also, I can try adding that argument here and see if that fixes the issue.

wyardley:
> Also, I can try adding that argument here and see if that fixes the issue.

So, just FWIW, I checked and adding `--config=pyproject.toml` to the args for pre-commit config doesn't seem to change the behavior (also, I noticed that ruff does seem to pick up settings from `pyproject.toml` even when it's not called with that argument).  If it matters, I'm also not seeing codes for the specific findings (for example, `ERA001`) ignored in the config file.

Playing around with it locally, I _think_ the actual issue is that having the `--diff` flag set seems to suppress some of the findings. If you try `ruff check --fix` vs `ruff check --fix --diff` 

```shell
$ ruff check --fix
[...]
tests/unit/semantic_release/hvcs/test_gitlab.py:
  612:5 SIM117 Use a single `with` statement with multiple contexts instead of nested `with` statements

tests/util.py:
  15:34 N812 Lowercase `config` imported as non-lowercase `cliConfigModule`

Found 26 errors.
$ ruff check --diff
$ echo $?
0
```

So even though there are findings, `ruff check --diff` seems to be exiting 0 in any case (note: I think it's not printing anything else because there are no _fixable_ findings, even though there are findings).

That said, we can leave this as it is here, and I can try ignoring and / or chipping away some of those other findings in another PR.

wyardley:
I'm trying to fix the mypy pre-commit issue with `--all-files` too, but may have to figure that out separately. It basically has to do with the way files are provided / passed 

Just setting `semantic-release/` as a final `args` value doesn't seem to do it properly, probably because it expects filenames to be passed in last as a positional argument, but with `pre-commit run --all-files`, it's throwing at least one error

```
% mypy semantic_release                                                         
Success: no issues found in 46 source files
% python3 -m mypy --config pyproject.toml --ignore-missing-imports semantic_release
Success: no issues found in 46 source files
% pre-commit run --all-files
[...]
semantic_release/cli/commands/publish.py:66:9: error: "HvcsBase" has no
attribute "upload_dists"  [attr-defined]
            hvcs_client.upload_dists(tag=tag, dist_glob=pattern)
            ^~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 46 source files)
```

Another thing I can do separately, but I _think_ with the versions of `mypy` and `ruff` we have, we may not need to set the config file anymore for either?

wyardley:
Threw up a draft first pass in #961 at the rest of the ruff failures and removing `--diff` from the CI job

codejedi365:
> % python3 -m mypy --config pyproject.toml --ignore-missing-imports semantic_release
> Success: no issues found in 46 source files
> % pre-commit run --all-files
> [...]
> semantic_release/cli/commands/publish.py:66:9: error: "HvcsBase" has no
> attribute "upload_dists"  [attr-defined]
>             hvcs_client.upload_dists(tag=tag, dist_glob=pattern)
>             ^~~~~~~~~~~~~~~~~~~~~~~~
> Found 1 error in 1 file (checked 46 source files)
> ```

Let's just throw a `# type: ignore` on this because this will be refactored soon anyway. 

> Another thing I can do separately, but I _think_ with the versions of `mypy` and `ruff` we have, we may not need to set the config file anymore for either?

Probably but not sure. The less command line arguments the better.

wyardley:
> Probably but not sure. The less command line arguments the better.

Ok, see how you think this looks now - removed what I think should be unnecessary args, also added that ignore.

codejedi365 (approved):

codejedi365:
Thanks 🎉

PR #958, opened 2024-06-15
refactor: ruff fixes and remove / reflow commented code

wyardley:
- Remove apparently redundant literal union (https://docs.astral.sh/ruff/rules/redundant-literal-union/)
- Switch to list comprehension in `tests/fixtures/git_repo.py`
- Shift around or remove some commented code to make `ruff` happy.

All the changes are test related, so if you want `test` vs `refactor`, let me know, or feel free to adjust the type when merging.


codejedi365:
From some of the changes you are adding here, I think ruff is not reading our configuration. Or if it is than your version of pre-commit does not match the ruff version in pyproject.toml. The reason I say this is that in the lint job it is running ruff but it hasn't been an issue lately. The `...` in the type definitions is one of the things that ruff has gone back and forth with. First it was on a new line and a few versions back I had to move it all to inline.

wyardley:
Let me make sure I have the right version - should have installed the one in the project config, but maybe there’s a discrepancy somewhere?

wyardley:
FWIW:
```
% ruff --version
ruff 0.4.4
```

https://github.com/python-semantic-release/python-semantic-release/blob/67971bf454b8ac56345cc51e939a00aa391d9117/pyproject.toml#L77

wyardley:
Maybe the updates in #957 will help? I notice that it's got the formatting one pinned back, though maybe now the two can be consolidated?

Either way, I think the more usual default of the `ruff` pre-commit hook is to only operate on the changed files vs. all files as it's currently configured.

wyardley:
@codejedi365 see if you like this better - got rid of all the autoformatting. If these two don't make sense, also feel free to just close.

codejedi365 (approved):
Thanks for all the work. Not at my computer to try the ruff version changes, but thanks for adapting.

codejedi365:
I agree it should operate on only the changed files

PR #961, opened 2024-06-17
chore: re-enable ruff warnings

wyardley:
Having the `--diff` flag in `ruff check` seems to suppress failures from happening in CI. Remove the `--diff` flag, and then ignore or adjust to resolve the remaining `ruff` failures. This will help prevent new ruff errors from going unnoticed and getting checked in, at the expense of a little bit of clutter in the code.

Note: this can be merged with the two atomic commits, or squashed as one.

We should probably do this after #957 goes in (may require a little conflict resolution on one file)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -222,6 +222,8 @@ ignore = [
   "DTZ",
   # flake8-errmsg
   "EM",
+  # Some todos and some examples; leave this disabled for now
+  "ERA001",
   # don't compare types, use isinstance()
   # sometimes using type(x) == y is deliberately chosen to exclude
   # subclasses
--- a/semantic_release/cli/commands/cli_context.py
+++ b/semantic_release/cli/commands/cli_context.py
@@ -58,7 +58,7 @@ class CliContextObj:
 
         try:
             if was_conf_file_user_provided and not conf_file_exists:
-                raise FileNotFoundError(
+                raise FileNotFoundError(  # noqa: TRY301
                     f"File {self.global_opts.config_file} does not exist"
                 )
 
--- a/semantic_release/cli/config.py
+++ b/semantic_release/cli/config.py
@@ -145,9 +145,12 @@ class RemoteConfig(BaseModel):
     @model_validator(mode="after")
     def set_default_token(self) -> Self:
         # Set the default token name for the given VCS when no user input is given
-        if not self.token and self.type in _known_hvcs:
-            if env_token := self._get_default_token():
-                self.token = env_token
+        if self.token:
+            return self
+        if self.type not in _known_hvcs:
+            return self
+        if env_token := self._get_default_token():
+            self.token = env_token
         return self
 
     def _get_default_token(self) -> str | None:
@@ -406,7 +409,7 @@ class RuntimeContext:
         return masker
 
     @classmethod
-    def from_raw_config(
+    def from_raw_config(  # noqa: C901
         cls, raw: RawConfig, global_cli_options: GlobalCommandLineOptions
     ) -> RuntimeContext:
         ##
--- a/semantic_release/hvcs/bitbucket.py
+++ b/semantic_release/hvcs/bitbucket.py
@@ -59,7 +59,7 @@ class Bitbucket(RemoteHvcsBase):
         hvcs_api_domain: str | None = None,
         token: str | None = None,
         allow_insecure: bool = False,
-        **kwargs: Any,
+        **kwargs: Any,  # noqa: ARG002
     ) -> None:
         super().__init__(remote_url)
         self.token = token
--- a/semantic_release/hvcs/remote_hvcs_base.py
+++ b/semantic_release/hvcs/remote_hvcs_base.py
@@ -34,7 +34,7 @@ class RemoteHvcsBase(HvcsBase, metaclass=ABCMeta):
 
     DEFAULT_ENV_TOKEN_NAME = "HVCS_TOKEN"  # noqa: S105
 
-    def __init__(self, remote_url: str, *args: Any, **kwargs: Any) -> None:
+    def __init__(self, remote_url: str, *_args: Any, **_kwargs: Any) -> None:
         super().__init__(remote_url)
         self._hvcs_domain: Url | None = None
         self._api_url: Url | None = None
--- a/tests/command_line/conftest.py
+++ b/tests/command_line/conftest.py
@@ -9,7 +9,7 @@ import pytest
 from click.testing import CliRunner
 from requests_mock import ANY
 
-from semantic_release.cli import config as CliConfigModule
+from semantic_release.cli import config as cli_config_module
 from semantic_release.cli.config import (
     GlobalCommandLineOptions,
     RawConfig,
@@ -57,7 +57,7 @@ def mocked_git_push(monkeypatch: MonkeyPatch) -> MagicMock:
     """Mock the `Repo.git.push()` method in `semantic_release.cli.main`."""
     mocked_push = MagicMock()
     cls = prepare_mocked_git_command_wrapper_type(push=mocked_push)
-    monkeypatch.setattr(CliConfigModule.Repo, "GitCommandWrapperType", cls)
+    monkeypatch.setattr(cli_config_module.Repo, "GitCommandWrapperType", cls)
     return mocked_push
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,7 +7,7 @@ from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING
 
 import pytest
-from git import Commit
+from git import Commit, Repo
 
 from tests.fixtures import *
 from tests.util import remove_dir_tree
@@ -57,7 +57,7 @@ def netrc_file(
     exception = None
     try:
         yield _netrc_file
-    except Exception as err:
+    except Exception as err:  # noqa: BLE001
         exception = err
     finally:
         for context_manager in entered_context_managers:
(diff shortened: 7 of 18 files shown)


pyproject.toml
```diff
@@ -222,6 +222,8 @@ ignore = [
   "DTZ",
   # flake8-errmsg
   "EM",
+  # Some todos and some examples; leave this disabled for now
+  "ERA001",
```
wyardley, line 226: already resolved some of these in #958

semantic_release/cli/commands/cli_context.py
```diff
@@ -58,7 +58,7 @@ def _init_runtime_ctx(self) -> RuntimeContext:
 
         try:
             if was_conf_file_user_provided and not conf_file_exists:
-                raise FileNotFoundError(
+                raise FileNotFoundError(  # noqa: TRY301
```
wyardley, line 61: @codejedi365: it _looked_ to me like this could maybe go outside the try / except, but didn't want to mess with it here?

semantic_release/cli/config.py
```diff
@@ -145,7 +145,7 @@ def resolve_env_vars(cls, val: Any) -> str | None:
     @model_validator(mode="after")
     def set_default_token(self) -> Self:
         # Set the default token name for the given VCS when no user input is given
-        if not self.token and self.type in _known_hvcs:
+        if not self.token and self.type in _known_hvcs:  # noqa: SIM102
```
wyardley, line 148: I think this reads cleaner on two lines, so left this and the other vs. resolving, but we can do on a single line if you'd like.

semantic_release/hvcs/bitbucket.py
```diff
@@ -59,7 +59,7 @@ def __init__(
         hvcs_api_domain: str | None = None,
         token: str | None = None,
         allow_insecure: bool = False,
-        **kwargs: Any,
+        **kwargs: Any,  # noqa: ARG002
```
wyardley, line 62: I'm not sure - seems like we maybe don't need kwargs on either the parent or child for these? But not sure if I'm missing some implicit way they're used?

semantic_release/hvcs/remote_hvcs_base.py
```diff
@@ -34,7 +34,7 @@ class RemoteHvcsBase(HvcsBase, metaclass=ABCMeta):
 
     DEFAULT_ENV_TOKEN_NAME = "HVCS_TOKEN"  # noqa: S105
 
-    def __init__(self, remote_url: str, *args: Any, **kwargs: Any) -> None:
+    def __init__(self, remote_url: str, *args: Any, **kwargs: Any) -> None:  # noqa: ARG002
```
wyardley, line 37: If you think we can remove `**kwargs` here, I think we could also use it on the various specific vcs systems that inherit?

I don't think just renaming to `**_kwargs` is the right fix, but that would also shut up ruff, I think?

tests/command_line/conftest.py
```diff
@@ -9,7 +9,7 @@
 from click.testing import CliRunner
 from requests_mock import ANY
 
-from semantic_release.cli import config as CliConfigModule
+from semantic_release.cli import config as CliConfigModule  # noqa: N812
```
wyardley, line 12: Wasn't sure if this needs to be pascal case here, or if I we can just do `cliconfigmodule` or `cli_config_module` and update the couple of places where it's used?

tests/conftest.py
```diff
@@ -93,6 +93,6 @@ def _teardown_cached_dir(directory: Path | str) -> Path:
 @pytest.fixture(scope="session")
 def make_commit_obj() -> MakeCommitObjFn:
     def _make_commit(message: str) -> Commit:
-        return Commit(repo=Repo(), binsha=Commit.NULL_BIN_SHA, message=message)
+        return Commit(repo=Repo(), binsha=Commit.NULL_BIN_SHA, message=message)  # noqa: F405
```
wyardley, line 96: wasn't sure about this one, not sure if there's a better fix vs. ignoring?

tests/util.py
```diff
@@ -12,7 +12,7 @@
 
 from semantic_release.changelog.context import make_changelog_context
 from semantic_release.changelog.release_history import ReleaseHistory
-from semantic_release.cli import config as cliConfigModule
+from semantic_release.cli import config as cliConfigModule  # noqa: N812
```
wyardley, line 15: same feedback as other tests file - could rename it if there's no reason it has to be camel case here.

semantic_release/cli/commands/cli_context.py
codejedi365: No, its purposefully inside the try/except so that the `click` context is properly exited cleanly.

semantic_release/cli/config.py
codejedi365: Leave it as the single line, because if you make it 2 this entire thing becomes 3 nested ifs which is more hideous.

codejedi365, line 148:
```suggestion
        if self.token:
            return self
        if self.type not in _known_hvcs:
            return self
        if env_token := self._get_default_token():
            self.token = env_token
        return self
```

codejedi365: This is probably what it should of been in the first place, following never nesting practice.

semantic_release/hvcs/bitbucket.py
codejedi365: Its for the next improvement. It doesn't have much value currently but it will when different HVCS constructors have different parameters.  You must use the `**kwargs` argument to consume any extra args that are not recognized and gracefully ignore them especially because the constructor signature will be slightly different.

tests/unit/semantic_release/hvcs/test_gitlab.py
```diff
@@ -609,7 +609,7 @@ def test_create_or_update_release_when_create_fails_and_update_fails(
     )
 
     # Execute in mocked environment expecting a GitlabUpdateError to be raised
-    with create_release_patch, edit_release_notes_patch, get_release_by_id_patch:
+    with create_release_patch, edit_release_notes_patch, get_release_by_id_patch:  # noqa: SIM117
```
codejedi365, line 612: What is this about... seems like a `ruff` bug. gahh i hate using <1.0 released software.

tests/conftest.py
codejedi365: I don't even know what is wrong with this line.

F405: https://docs.astral.sh/ruff/rules/undefined-local-with-import-star-usage/

semantic_release/hvcs/remote_hvcs_base.py
codejedi365: This is a dumb error from ruff. yes it is needed as stated above, I'm fine with using the underscore rather than a bunch of ignores.

codejedi365 (commented):
see comments above.

tests/unit/semantic_release/hvcs/test_gitlab.py
wyardley: It's the nested `with`; the alternative they suggest is to do `with A() as a, B() as b:`, but of course that would look pretty messy here:

https://docs.astral.sh/ruff/rules/multiple-with-statements/

If the raises could be a decorator, maybe that would look better, but I think just ignoring it is probably fine here?

tests/conftest.py
wyardley: The full error is:
```
F405 `Repo` may be undefined, or defined from star imports
```

I think because `Repo()` isn't imported / defined directly in `conftest.py`? Is it getting it implicitly from somewhere else (another import)?

codejedi365: Ah ok, yeah that is it. Just add to line 10, `from git import Commit, Repo`. It was not intended result

codejedi365 (commented):
Looking better

tests/conftest.py
wyardley: Yeah, wasn't sure if it was just really a missing import. Thanks for the heads up - added that in and removed the ignore.

tests/unit/semantic_release/hvcs/test_gitlab.py
codejedi365: Well the doc doesn't consider this situation for creation of the context manger is multiple lines and prior to the with statement.

Yes, ignore is fine here.

codejedi365 (approved):

wyardley:
just noting here, I poked at this some more to create the bug report upstream, and the behavior does _kind of_ make sense... it's just confusing
`ruff --diff` will exit 0 if there are no diffs (proposed changes) reported and exit 1 if there are. However, there are errors / findings that do not, or cannot, generate changes to files, but are still errors.

It may be worth adding `ruff --diff` as part of the pipeline somewhere to display the changes (if we want) - however, using it the way we were originally probably doesn't make sense.

codejedi365:
Thanks for looking into this, definitely is a weird result. Either way all I want is it to display where the error is located. The diff is a nice to have where it shows what to fix but I can also look up the rule for where it indicates an error exists.

wyardley:
Yeah, re-reading the docs for `--diff` and for `--fix-only` (which it notes is implied), it actually does make sense:
```
      --diff
          Avoid writing any fixed files back; instead, output a diff for each changed
          file to stdout. Implies `--fix-only`
[...]
      --fix-only
          Apply fixes to resolve lint violations, but don't report on leftover
          violations. Implies `--fix`. Use `--no-fix-only` to disable or `--unsafe-fixes`
          to include unsafe fixes
```

I think in this case "report on" also implies the exit status as well, in other words, because `--diff` implies `--fix-only`, it is expected that it doesn't report non-fixable findings _and_ exits 0 for non-fixable findings. So the interaction of flags is a little unexpected.

I agree that things are fine as it is now. It would be nice to have a way to both report the fixable changes as a diff, and also check for findings, with a single command, but seems like not possible as-is. Short of some additional notes in the docs, or adding some additional flags, might be difficult to fix as well.

PR #963, opened 2024-06-19
feat(parser-emoji): ability to define other emoji

justin-pierce:
I didn't like how using emojis that didn't trigger a release would dump them all together in a generic "Other" category in release notes. 

This simply lets you define `other_emoji` so they appear under their own header. 

Side note: is there a reason the emoji id (like `:construction_worker:`) is used instead of the actual emoji (like 👷) in the default configuration? In my personal configurations I just use the actual emojis and it seems to work fine and I like that it renders the emoji in my IDE.

### DIFF
--- a/docs/commit-parsing.rst
+++ b/docs/commit-parsing.rst
@@ -116,6 +116,11 @@ The default configuration options for
         ":robot:",
         ":green_apple:",
     ]
+    other_tags = [
+        ":pencil:",
+        ":construction_worker:",
+        ":recycle:",
+    ]
 
 .. _commit-parser-scipy:
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -657,6 +657,9 @@ the expections from ``commit_parser`` value to default options value.
                                  ":apple:", ":penguin:", ":checkered_flag:", ":robot:",
                                  ":green_apple:"
                              ]
+                             other_tags = [
+                                 ":pencil:", ":construction_worker:", ":recycle:"
+                             ]
 
 ``"scipy"``         ->   .. code-block:: toml
 
--- a/semantic_release/commit_parser/emoji.py
+++ b/semantic_release/commit_parser/emoji.py
@@ -41,6 +41,11 @@ class EmojiParserOptions(ParserOptions):
         ":robot:",
         ":green_apple:",
     )
+    other_tags: Tuple[str, ...] = (
+        ":pencil:",
+        ":construction_worker:",
+        ":recycle:",
+    )
     default_bump_level: LevelBump = LevelBump.NO_RELEASE
 
 
@@ -66,7 +71,7 @@ class EmojiCommitParser(CommitParser[ParseResult, EmojiParserOptions]):
 
     def parse(self, commit: Commit) -> ParseResult:
         all_emojis = (
-            self.options.major_tags + self.options.minor_tags + self.options.patch_tags
+            self.options.major_tags + self.options.minor_tags + self.options.patch_tags + self.options.other_tags
         )
 
         message = str(commit.message)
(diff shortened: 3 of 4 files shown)


wyardley (commented):
I like the idea; might even be good to bake in some common non-triggering types as defaults (even for non-triggering types like docs / ci / chore)?

`other_tags` seems fine; maybe `non_triggering_tags` or something similar could also work?

I guess https://github.com/carloscuesta/gitmoji/blob/master/packages/gitmojis/src/gitmojis.json is the closest thing to a standard here?

wyardley:
> Side note: is there a reason the emoji id (like :construction_worker:) is used instead of the actual emoji

Just commenting as a user here, but I do personally think having `:foo:` vs the actual emoji just seems safer to me somehow 🤷

justin-pierce:
> I like the idea; might even be good to bake in some common non-triggering types as defaults (even for non-triggering types like docs / ci / chore)?

Yeah for defaults I (at least think I) added:
- `:pencil:` for docs (I use 📝`:memo:` but the tests here already mentioned `:pencil:` relating to docs so I kept it)
- `:construction_worker:` for cicd
- `:recycle:` for refactoring

Wasn't sure how far to take the defaults. I like gitmoji for reference (and used it for defaults other than what's noted) but I personally deviate a little and think simpler is better, as there can be some overlap/ambiguity when there's too many (and at some point it's annoying to remember them all when they get super specific). 

> `other_tags` seems fine; maybe `non_triggering_tags` or something similar could also work?

Yeah wasn't sure what to call it. I looked at how the other parsers worked but I preferred how the emoji parser didn't double up with an `all` list. `non_triggering_tags` also makes sense. 

> Just commenting as a user here, but I do personally think having :foo: vs the actual emoji just seems safer to me somehow 🤷. Seems like it could be less likely for problems to come up around encoding or viewing the files at any rate.

Yeah figured it was something along those lines -- just wanted to make sure I wasn't missing any specific risks when I use emoji themselves, as they seem to render everywhere I need them to (whereas the codes don't render in pycharm, for example). Easy enough to customize so I don't think it's an issue, was just curious. 👍

justin-pierce:
went ahead and swapped `:pencil:` with `:memo:` and changed `other_tags` to `non_triggering_tags`

codejedi365:
Sorry for the delay, been traveling, hopefully get to the review this week. 

Initially, I'm not seeing any testing changes so have you considered how to validate your fix and how to prevent it from regression later on?

justin-pierce:
> Sorry for the delay, been traveling, hopefully get to the review this week.
> 
> Initially, I'm not seeing ~any~ (there are some adjustments to constants) testing changes related to the changelog generation you were hoping for so have you considered how to validate your fix and how to prevent it from regression later on? I do realize some of the changelog testing is complicated.

Yeah was tough for me to parse the testing structure. I did edit these:

```
        # No release with specified emoji
        (
            ":memo: Documentation changes",
            LevelBump.NO_RELEASE,
            ":memo:",
            [":memo: Documentation changes"],
            [],
        ),
        # No release with random emoji
        (
            ":construction: Work in progress",
            LevelBump.NO_RELEASE,
            "Other",
            [":construction: Work in progress"],
            [],
        ),
```

My understanding is that this does test a commit message's effect on both the type of release and how it appears in changelogs. The first tests to make sure :memo: puts it under the :memo: header because :memo: is defined in the defaults. The second uses :construction:, which is not in the defaults, so it tests to confirm it's under the Other header. 

Sounds like I'm missing something though?

codejedi365:
I'm sorry, this was my fault, I was extremely busy the past two months and I didn't get back to reviewing this PR a while back. I will review more over the weekend. I appreciate your contribution to the project.

codejedi365:
Refactored and merged into default branch via #1410 since I can't edit this PR.

PR #964, opened 2024-06-19
chore: switch to ruff doc formatting and add ignore types

wyardley:
Here's another small one; let me know what you think:

- Remove `blacken-docs` in pre-commit config.
- Set `docstring-code-format` to `true` for ruff.
- Resolve `blanket-type-ignore` pre-commit check by adding types to mypy type ignores

Basically, [this](https://github.com/astral-sh/ruff/pull/8854) replaces what `blacken-docs` would have done.

Note: if you don't care about the type ignores, I can remove that check (`python-check-blanket-type-ignore` from https://github.com/pre-commit/pygrep-hooks, with or without reverting those annotations). Note: that check will run for users who have pre-commit setup / installed, but not for other users, so it's not enforced in CI, which is probably part of how these 3 ended up here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -287,9 +287,10 @@ external = ["V"]
 task-tags = ["NOTE", "TODO", "FIXME", "XXX"]
 
 [tool.ruff.format]
-quote-style = "double"
+docstring-code-format = true
 indent-style = "space"
 line-ending = "lf"
+quote-style = "double"
 
 [tool.ruff.lint.per-file-ignores]
 # Imported but unused
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,12 +29,6 @@ repos:
       - id: mixed-line-ending
       - id: check-ast
 
-  - repo: https://github.com/adamchainz/blacken-docs
-    rev: 1.16.0
-    hooks:
-      - id: blacken-docs
-        additional_dependencies: ["black==23.10.1"]
-
   - repo: https://github.com/asottile/pyupgrade
     rev: v3.16.0
     hooks:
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -287,9 +287,10 @@ external = ["V"]
 task-tags = ["NOTE", "TODO", "FIXME", "XXX"]
 
 [tool.ruff.format]
-quote-style = "double"
+docstring-code-format = true
 indent-style = "space"
 line-ending = "lf"
+quote-style = "double"
```
wyardley, line 293: Note: the rest of these settings (quote-style, line-ending, indent-style) seem to mostly / all be default. @codejedi365: if you prefer, I can pull them out; in my testing, removing them seems to be a noop.

codejedi365:
Sorry for the delay, been traveling, I hope to review your PR this week.

pyproject.toml
codejedi365: In this case, I'd rather just leave them as explicit defaults, especially since Ruff is so immature of a package that anything could change.

codejedi365 (approved):
looks good, thanks!

codejedi365:
> Note: if you don't care about the type ignores, I can remove that check (`python-check-blanket-type-ignore` from https://github.com/pre-commit/pygrep-hooks, with or without reverting those annotations). Note: that check will run for users who have pre-commit setup / installed, but not for other users, so it's not enforced in CI, which is probably part of how these 3 ended up here.

I would prefer explicit type ignores that say what rule they are ignoring. So the check is the right one. We should find a way to apply the rule in the CI somehow.

PR #971, opened 2024-07-04
chore: set `ignore-without-code` for mypy

wyardley:
As discussed in https://github.com/python-semantic-release/python-semantic-release/pull/964#issuecomment-2208080852, this will check for explicit error types in type ignores.

- Set `ignore-without-code` in mypy settings 

    - https://mypy.readthedocs.io/en/stable/error_codes.html#requiring-error-codes
    - https://docs.astral.sh/ruff/rules/blanket-type-ignore/

- Remove `pygrep` rule in pre-commit config to avoid flagging it twice locally (since there's a mypy pre-commit config already)

Tested locally that this should flag the case where the error type is not set.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -237,6 +237,8 @@ ignore = [
   "INP001",
   # Errors should end with "Error"
   "N818",
+  # mypy prevents blanket-type-ignore
+  "PGH003",
   # Fixtures that do not return a value need an underscore prefix.  The rule
   # does not handle generators.
   "PT004",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,6 +156,7 @@ show_error_context = true
 pretty = true
 error_summary = true
 follow_imports = "normal"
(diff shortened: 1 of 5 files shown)


codejedi365:
@wyardley, added two commits on, one to disable `ruff`'s tracking of type errors so we don't have duplicate error messages, and then removed the `noqa` entry that is no longer applicable

pyproject.toml
```diff
@@ -236,6 +237,8 @@ ignore = [
   "INP001",
   # Errors should end with "Error"
   "N818",
+  # mypy prevents blanket-type-ignore
+  "PGH003",
```
wyardley, line 241: Oh interesting. I was wondering why this hadn’t fired on the ruff side of things. Thought I had seen some before that weren’t ruff ignited but maybe not. So I guess we could have left the mypy config as is and removed the ruff ignores 🤷 but either way this works.

PR #1002, opened 2024-08-17
build(deps-dev): update dev dependencies and pre-commit hooks

wyardley:
## Purpose
- Update ruff to v0.6.3
- Update mypy to v1.11.2
- Update some pre-commit hooks

Looked at updating `pre-commit` itself in dev dependencies, however, looks like Python 3.8 (which checks still run on) doesn't support anything after ~ 3.5.

#1009  and #1006 would likely auto-close after this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
 
   # Linters and validation
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.0
+    rev: v0.6.3
     hooks:
       - id: ruff
         name: ruff (lint)
@@ -50,7 +50,7 @@ repos:
         name: ruff (format)
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: "v1.10.0"
+    rev: "v1.11.2"
     hooks:
       - id: mypy
         additional_dependencies:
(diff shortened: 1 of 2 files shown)


wyardley:
@codejedi365 not time-sensitive, but does this one look reasonable to you?

codejedi365:
Hey @wyardley, sorry for the delay, had a lot of changes in my life recently and have not had much time over the past month. 

The changes look fine. It's just lower on the priority as it's dev env stuff and I've been wanting to sit down and review the changelogs of the dependencies before introduction. In the grand scheme of things, I would rather that each dependency is bumped in their own commits so that they are distinguishable in the git history in case one causes unforeseen issues. I also already have the dependabot PRs open and would rather add the "pre-commit" update commit added to those branches. So they are in sync. 

Does that make sense? Note you can open PRs against other branches not just master.

One hesitancy I have with the constant update of ruff is that it's still not above a `1.0` which means it could have breaking changes which causes more work than it's worth every week when I have bigger features/issues to address. With each bump new rules are added and now I have to update code accordingly with current branches I'm already working and haven't merged in. 

I would appreciate it if you desire to make more contributions to more to start tackling some of those "good first issue" tagged issues as I'm really trying to knock down the ticket queue below 70!

codejedi365:
> Yes, it’s remembering to bump them in sync in pre-commit that’s a pain but that totally makes sense (renovate can do it fwiw)

It's a pain but I don't really forget that stuff. I've looked into renovate and looks good but have not implemented it yet.

> I think ruff should error in CI now if there are breaking changes that affect the existing code.

That's why I have locked in the specific version of ruff (Andy mypy). I don't want an arbitrary bump in version because of flexible pip resolution to cause failures in CI that were not a problem in a previous run.

PR #1012, opened 2024-08-31
Add script to build image; then use the pre-built image for the GH action

pranavmishra90:


pranavmishra90:
Apologies, i was testing this action, but pulled to the upstream

Issue #1018, opened 2024-09-03
Version variable is not changed when using PEP440 specification


lukasbehammer:
### Description
If the version variable includes a pre-release segment , post-release segment or development segment (PEP440 / PyPA Version Specifiers), the version is not getting updated.

### Expected behavior
Version variable gets updated to new version if a release is made.

### Actual behavior
Version variable stays the same and does not get updated if a release is made.

### Environment

-  **Operating System (w/ version):** Windows 11 (local), Ubuntu (GitHub Actions Runner)
-  **Python version:** 3.11.9
-  **Pip version:** 24.0
-  **Semantic-release version:** 9.8.6
-  **Build tool (w/ version):** meson-python 0.16.0

<details>
<summary><code>pip freeze</code></summary>

```log
alabaster==0.7.16
annotated-types==0.7.0
anyio==4.3.0
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==2.4.1
async-lru==2.0.4
attrs==23.2.0
Babel==2.15.0
backports.tarfile==1.2.0
beautifulsoup4==4.12.3
bleach==6.1.0
build==1.2.1
cachetools==5.3.3
certifi==2024.2.2
cffi==1.16.0
cfgv==3.4.0
chardet==5.2.0
charset-normalizer==3.3.2
click==8.1.7
click-option-group==0.5.6
colorama==0.4.6
comm==0.2.2
contourpy==1.2.1
coverage==7.5.1
cycler==0.12.1
debugpy==1.8.1
decorator==5.1.1
defusedxml==0.7.1
distlib==0.3.8
docutils==0.21.2
dotty-dict==1.3.1
executing==2.0.1
fastjsonschema==2.19.1
filelock==3.14.0
fonttools==4.51.0
fqdn==1.5.1
fsspec==2024.5.0
furo==2024.5.6
gitdb==4.0.11
GitPython==3.1.43
h11==0.14.0
httpcore==1.0.5
httpx==0.27.0
identify==2.5.36
idna==3.7
imageio==2.34.1
imagesize==1.4.1
importlib_metadata==8.0.0
importlib_resources==6.4.0
iniconfig==2.0.0
intel-openmp==2021.4.0
ipykernel==6.29.4
ipython==8.24.0
ipywidgets==8.1.2
isoduration==20.11.0
jaraco.classes==3.4.0
jaraco.context==5.3.0
[...]

Issue #1000, opened 2024-08-10
Issues with "non-canonical" dev version


david-waterworth:
## Question

I'm running into issues with python-semantic-release and poetry with my dev builds. I probably have some config set wrong, but `python-semantic-release` has generated `version = "0.8.0-dev.2"` in my `pyproject.toml` and when I run `poetry build` I end up with

```
Building my-project (0.8.0-dev.2)
  - Building sdist
  - Built my-project-0.8.0.dev2.tar.gz
  - Building wheel
  - Built my-project-0.8.0.dev2-py3-none-any.whl
``` 

Note there's a `.` (period) between the dev token and the number 2 in the python-semantic-release generated version string but not in the filename. This is causing me grief because in my build script I'm using `poetry version` to find the file that was created, and I cannot because they don't match.

It looks to me like `python-semantic-release` is generating a version that matches the longer regex from https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions but it doesn't match the "canonical" version (`r'^([1-9][0-9]*!)?(0|[1-9][0-9]*)(\.(0|[1-9][0-9]*))*((a|b|rc)(0|[1-9][0-9]*))?(\.post(0|[1-9][0-9]*))?(\.dev(0|[1-9][0-9]*))?$'`)

https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers also states

```
The canonical public version identifiers MUST comply with the following scheme:

[N!]N(.N)*[{a|b|rc}N][.postN][.devN]
```

Which isn't true for `0.8.0-dev.2` (should be `0.8.0.dev2` if I'm reading this correctly).

I'm assuming that `poetry build` "canonicalizes" the version when creating the sdist/wheel files?

Is it possible to ensure that `python-semantic-release` only generates "canonical" version strings in the first place? Or am I missing something?

### Configuration

```
[tool.semantic_release]
upload_to_pypi = false
upload_to_vcs_release = false
build_command = "./scripts/build.sh && ./scripts/publish.sh"
tag_format = "v{version}"
version_toml = [
    "pyproject.toml:tool.poetry.version"
]

[tool.semantic_release.branches.main]
match = "(main|master)"
prerelease_token = "rc"
prerelease = false

[tool.semantic_release.branches.other]
match = "^(feat|fix|perf)/.+"
prerelease_token = "dev"
prerelease = true
```

codejedi365:
[...]

Issue #455, opened 2022-06-09
Support PEP 440 versioning


thearchitector:
## Description

It would be useful to support [PEP 440](https://peps.python.org/pep-0440/) versioning alongside, or instead of, semantic versioning (they are similar, with subtle differences in identifiers).

## Use cases

Poetry enforces PEP 440. `pip`, `pypi` and `setuptools` all will or currently do support it, and may also require it. Since this package is intended for python specifically, it would make sense to support versioning that adheres to the accepted format.

## Possible implementation

Currently, there is a `--prerelease` option. In PEP 440, you can also have post releases, dev releases, and local versions. It would be nice / required to have those options as kwarg flags as well (like `--postrelease`, `--dev`, `--local`).

In terms of `prerelease`, the `prerelease_tag` would have to go from accepting any input to accepting only those valid in the spec (`alpha`, `beta`, `pre`, `preview`, `a`, `b`, `c`, and `rc`).

PEP 440 also provides [regular expressions](https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions) that can be used to check and extract existing version identifiers, which could be useful in automatically determining the next version of a valid release.

bernardcooke53:
@thearchitector would this requirement be met by being able to configure `prerelease_tag` on a per-branch basis? It sounds like you want to create different prerelease formats based on (I'm guessing) the branch you're using?

I don't think restricting to PEP 440 versions only is a good idea - although PSR is a Python tool, nothing precludes it from running against projects in other languages. For example it would be a small leap to support config in `Cargo.toml` for Rust projects in the future - but Rust projects won't need to know/care about PEP 440

Aaryia:
I am finding the tool really useful in theory, but I am not able to use it for work projects due to the lack of support of PEP 440...
From what I gather, the format `X.Y.Z-tag.N` is not at all compatible with PEP 440, or at least is not considered canonical. It keeps us from publishing prerelease packages using PSR which is a shame.

A supported format would be `X.Y.ZtagN` where tag could be one of many, `rc`, `dev` or others, which can be covered with the use of `prerelease_tag`.

It would be nice to have at least a `--pep440` option or similar to natively support compatibility with PEP 440 prerelease versions.

bernardcooke53:
Ah yes - the "." before the revision specifier, sorry for not getting that sooner. I've hit the same pain before, I agree that a `--pep-440` would be valuable.

[...]

PR #1020, opened 2024-09-07
feat: implement basic pep440 support

lukasbehammer:
## Purpose

Adds PEP440 support as dicussed in #1000, #1018 and #455

## Rationale

I added version_compat as a cli flag and setting for branches. I also added PEP440 support, though not completely. Currently not added is support for multiple concurrent version specifiers (e.g. v1.2.3rc4.dev5) and support for epoch specifiers. Local versions have to be set manually via the buildmetadata option. Since the official regex was used, all possible pep440 version specifiers should at least be parsed from the tags. Setting the prerelease-token to one of the supported specifiers and setting as-prerelease makes a release with the token and prerelease version number 1. I have not yet implemented the auto-increase of the prerelease version. version_compat is not yet supported in the GitHub action. Docs are not yet updated.

## How did you test?

I tested manually with a small test [repository](https://github.com/lukasbehammer/python-semantic-release-pep440_test) and command line. I have not yet written tests for the new functionality.


lukasbehammer:
This is not yet ready to be merged. I'll have to write some tests first and some of the tests fail currently. @codejedi365 tell me what you think about this implementation and I'll go ahead writing tests, improving it and adding it to the Actions.

lukasbehammer:
Just a quick reminder @codejedi365 if you can have a quick look at this implementation of pep440 support, so that I know if I should go forward with the implementation.

semantic_release/version/translator.py
```diff
@@ -44,10 +41,12 @@ def __init__(
         self,
         tag_format: str = "v{version}",
         prerelease_token: str = "rc",  # noqa: S107
+        version_compat: str = "semver"
```
codejedi365, line 44: probably should be a enum as we have strictly 2 possibilities

codejedi365 (commented):
I agree with your approach. I'm glad you broke out a separate regex just for pep440 and I liked how you used the version translator as well.  

I do think where you have the type string for all the `version_compat` vars should be a defined enum as we strictly only have the 2.  Also is there a better name we can come up with?

My other thought is I'm not sure how much `version_compat` parameters we need throughout the code.  I would rather attempt to read semver then attempt to read pep404 when parsing the tag history (`tags_and_versions()`) and then convert that into a Version object with a flag set to if the tag was a pep440 or not.  That way we can actually support all tags (even in the odd chance that the user switched from one to the other).  Then like, you already did, adjust the `Version.__str__` method to print out a pep440 variant or a semver version depending on what configuration the user has set for what version types they want to use for the new versions. I would like to see it as a `--pep440` option to the `version` command, which would print any Version object to pep440 (as long as its compatible tokens).

Otherwise well done and make sure to add tests, update the documentation and add a GitHub action parameter.  If you need help with this, let me know. Also just as a heads up I have a few PRs I'm working simultaneously so there may be some conflicts as this PR likely will be quite large.  If it makes more sense, we should probably break it up into smaller doses and merge parts (given the functionality is not accessible to the user, ie. last thing to actually add is to enable the `--pep440` option).

codejedi365:
Sorry for the delay, I was sick and then out of town for a week. Thanks for the reminder.

codejedi365:
@lukasbehammer, did you have questions on my feedback?

lukasbehammer:
Ah sorry @codejedi365 . I haven't had much time lately. The earliest time I can do some more is probably around December/January. Though I'll try to fit some minor stuff, like tests etc. in such that it's easier to merge. If you have some time feel free to continue with it.

lukasbehammer:
I changed the existing tests to run with the new version-compat scheme, though I'm not sure how to implement the tests for the pep440 compatibility. Should  we copy the semver ones and run them again with pep440? What is the best approach?

codejedi365:
> Should  we copy the semver ones and run them again with pep440? What is the best approach?

For unit testing:

1. you will want to have a variant of the semver parse testing to ensure pep440 strings are parsed properly

2. You should probably add some comparator testing between a semver and pep440 parsed version

For system testing:

> Add to the `tests/command_line/` directory

1. Test the `--pep440` option that it generates a pep440 version the following repos variants

- all pep440 versions (this fixture doesn't exist, duplicate the repo with angular commits type, maybe a few of them since prereleases versions must be tested)

- mixed pep440 and semver versions  where someone started with pep440 and converted to semver

- mixed pep440 and semver versions  where someone started with semver and converted to pep440

2. Add the repo variants to the changelog_regenerate test

3. Add the new pep440 repos to any changelog update tests that make sense

codejedi365:
Remember to rebase soon as there have been some changes your branch conflicts with

lukasbehammer:
@codejedi365 regarding the --pep440 flag you suggested, I'd rather keep the --version-compat flag such that we're open to future versioning schemes.

lukasbehammer:
> Remember to rebase soon as there have been some changes your branch conflicts with

I'll do.

codejedi365:
> @codejedi365 regarding the --pep440 flag you suggested, I'd rather keep the --version-compat flag such that we're open to future versioning schemes.

Although I appreciate the forward look, I think the current name is less intuitive and I don't anticipate any more version schemes as the world has settled around semver fairly heavily. I know there is CalVer out there but that isn't directly compatible with our version bump construct. Any other version scheme would need to match the version bump construct. With that said, I stand by my comment to change to `--pep440` as the option. If my predictions is wrong, I'll deal with it when the time comes.

codejedi365:
Added another unit test idea to the list above.

lukasbehammer:
Is version_scheme better than version_compat? If so I'd refactor.

codejedi365:
@lukasbehammer, when you start running the tests you are probably going to want to add the `-nauto` flag for the xdist plugin to use multiple cpus. You probably will want to isolate the specific test files to run or test case while running locally because the whole suite takes a long time.

It's on the list of things to fix but it's not quite easy to change.

lukasbehammer:
Also I tried to implement the Enum. Didn't really work out. Maybe you @codejedi365 can have a look at it.

codejedi365:
Sure, I can take a look tomorrow

cfxegbert:
Should version_compat be a config file option instead of a command line option?  It would seem error prone to have to specify on the command line every time you bump the version.  It seems you would just want to configure it once in your pyproject.toml or releaserc.toml file.

src/semantic_release/version/version.py
```diff
@@ -167,11 +202,30 @@ def is_prerelease(self) -> bool:
...
     def __str__(self) -> str:
         full = f"{self.major}.{self.minor}.{self.patch}"
-        prerelease = (
-            f"-{self.prerelease_token}.{self.prerelease_revision}"
-            if self.prerelease_revision
-            else ""
-        )
+        if self.version_compat == "semver":
+            pre_str = f"-{self.prerelease_token}.{self.prerelease_revision}"
+        elif self.version_compat == "pep440":
```
cfxegbert, line 207: Since you are tranforming the prerelease_token in a lossy way the --print_tag command line option may not actually represent the actual tag that will be stored in the repo.  The print_tag takes the str(version) and converts it to a tag.  I would not not do the transformation and keep 'alpha', 'beta', ... since they are allowed by the pep440.

We should implement unit test for this case.

cfxegbert:
Current implementation has a hardcoded separator between the the release and pre/post/dev token and the separator between pre/post/dev and the number.  In pep440 "1.2.3-pre-1" is the same as "1.2.3pre.1".  Should we have a config option for the separators?  We could use the already existing prerelease option and allow something like
```toml
[semantic_release.branches.main]
match = "develop"
prerelease_token = "-dev."
prerelease = true
```
The above would generate 1.2.3-dev.1 and a `prerelease_token = "_dev"` would generate 1.2.3_dev1.  The checks on allowed prerelease_tokens could be done after the config file is parsed.  This would also easily handle the post release format of 1.2.3-1

We would just have to document it correctly.

src/semantic_release/version/version.py
cfxegbert: If this being done for normalization for sorting we should make a normalize method on version just for sorting.

codejedi365:
> Should version_compat be a config file option instead of both a command line option and a config file option? It would seem error prone to have to specify on the command line every time you bump the version. You would just want to configure it once in your pyproject.toml or releaserc.toml file. In my opinion version_compat should only be a config file option. It should not be both a cli and config option.

Yes, we should probably have a configuration option in the config for it as it likely is a lasting setting. I still would like to maintain the cli option as that is how we provide overrides to the github action. The github action is almost configureless as the defaults will load accordingly. We will not make pep440 the default so it will need to be a cli option.

src/semantic_release/version/version.py
codejedi365: > Since you are tranforming the prerelease_token in a lossy way the --print_tag command line option may not actually represent the actual tag that will be stored in the repo. The print_tag takes the str(version) and converts it to a tag. 

This is a good catch but actually unrelated to this PR. This is a bug and will be fixed separately. It is supposed to provide the actual tag that would match the repo. An e2e test will be written for this bug fix.

codejedi365: Yes the planned implementation is to evaluate the repo for all release tags and convert them to Version Objects. First we try the SEMVER parser and then secondly we try a pep440 parser to normalize all versions that match a tag format.  This way sorting is correct regardless of what version spec they used.  

The resulting tag or printed version (and/or tag) is where the `--pep440` parameter and a configuration option tweak PSR to output a pep440 formatted version.

codejedi365:
> Current implementation has a hardcoded separator between the the release and pre/post/dev token and the separator between pre/post/dev and the number. In pep440 "1.2.3-pre-1" is the same as "1.2.3pre.1". Should we have a config option for the separators? We could use the already existing prerelease option and allow something like
> 
> ```toml
> [semantic_release.branches.dev]
> match = "develop"
> prerelease_token = "-dev."
> prerelease = true
> ```
> 
> The above would generate 1.2.3-dev.1 and a `prerelease_token = "_dev"` would generate 1.2.3_dev1. The checks on allowed prerelease_tokens could be done after the config file is parsed. This would also easily handle the post release format of 1.2.3-1
> 
> We would just have to document it correctly.

I do not intend to provide configuration for the separator as that would result in non-conformative versions to the two supported standards.  The reason it is hardcoded is because PSR builds SemVer compliant versions and once this PR is complete, it will have the ability to create PEP440 compliant versions. I will have to check on those version similarities. I hope not... I just can't stand pep440.

cfxegbert:
> > Should version_compat be a config file option instead of both a command line option and a config file option? It would seem error prone to have to specify on the command line every time you bump the version. You would just want to configure it once in your pyproject.toml or releaserc.toml file. In my opinion version_compat should only be a config file option. It should not be both a cli and config option.
> 
> Yes, we should probably have a configuration option in the config for it as it likely is a lasting setting. I still would like to maintain the cli option as that is how we provide overrides to the github action. The github action is almost configureless as the defaults will load accordingly. We will not make pep440 the default so it will need to be a cli option.

Having a command line flag would be inconsistent with every other setting.  Why could the config file settings not work in a github action?

lukasbehammer:
> Having a command line flag would be inconsistent with every other setting.  Why could the config file settings not work in a github action?

I don't think the command line flag is inconsistent. All the other things can be either set by a config file (e.g. pyproject.toml) or by command line options. The GitHub action currently indeed is setup to use the cli options to call the code.

On another note: I still prefer the version_compat scheme, but have no obligations against using a pep440 flag and trying to evaluate the semver first.

codejedi365:
> Having a command line flag would be inconsistent with every other setting.  

That is a statement of opinion that I do not share.

> Why could the config file settings not work in a github action?

I did not say that. Configuration files are parsed in a GitHub Action. Cli options are for overrides. Just like in every other Linux tool.

src/semantic_release/version/version.py
cfxegbert: > This is a good catch but actually unrelated to this PR. This is a bug and will be fixed separately

That fix still converts the version to a string so this transformation when converting to a string will still break producing the next tag.  The only reason to normalize the version is for sorting.

codejedi365:
> Also I tried to implement the Enum. Didn't really work out. Maybe you @codejedi365 can have a look at it.

@lukasbehammer, sorry I didn't get back around to this, I've been working other fixes & features that have been lingering too long. Hopefully over the next month.

Issue #1023, opened 2024-09-13
NameError: name 'httpx' is not defined


kwevers:
### Description
Since today's [v4.11.0](https://github.com/python-gitlab/python-gitlab/releases/tag/v4.11.0) release of `python-gitlab` using the semantic release GitHub Actions fails with a `NameError: name 'httpx' is not defined` error.

### Expected behavior

Successful release creation

### Actual behavior

```
Run python-semantic-release/python-semantic-release@v9.8.8
  with:
    github_token: ***
    vcs_release: false
    root_options: -v
    directory: .
  env:
    COMMIT_AUTHOR: Kristof Wevers <kwevers@users.noreply.github.com>
[...]

PR #1024, opened 2024-09-13
build(deps): restrict `python-gitlab` dependency due to broken release

kwevers:
The `python-gitlab` package depends on `httpx` starting from v4.11.0

## Purpose

I believe this fixes https://github.com/python-semantic-release/python-semantic-release/issues/1023
Not sure if this is the right way to approach this though.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ dependencies = [
   "click ~= 8.0",
   "click-option-group ~= 0.5",
   "gitpython ~= 3.0",
+  "httpx ~=0.27",
   "requests ~= 2.25",
   "jinja2 ~= 3.1",
   "python-gitlab ~= 4.0",


pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
   "click ~= 8.0",
   "click-option-group ~= 0.5",
   "gitpython ~= 3.0",
+  "httpx ~=0.27",
   "requests ~= 2.25",
   "jinja2 ~= 3.1",
   "python-gitlab ~= 4.0",
```
codejedi365, line 31: If we were to make a change, we should instead prevent the broken release of `python-gitlab` from being selected. Utilizing the `! 4.11.0` syntax.

With this line 28 addition of httpx, is not required.

This is the recommended manner of fixing a broken dependency library.

codejedi365 (changes requested):
Thank you for the rapid engagement to resolve the issue in this project.  I don't think this situation should occur again since the package has been yanked but otherwise my recommendation above would be how this project would solve this issue.

PR #1035, opened 2024-09-27
ci(main-workflow): change `upload-to-gh-release` action to `publish-action@v9.9.0`

codejedi365:
## Purpose

- Update this project to be inline with our documentation
- Remove the deprecated action

## Rationale

Because we renamed the action we needed to update the implementation in our project as well.  As many people will use our project as a reference.  

Not merging this immediately until after the next release is also a slight test for making sure that our warning annotation works for `upload-to-gh-release` as we should see it in our Actions summary.  Then we will change it over to the new action version.

## How did you test?

Just considered that the name change doesn't need testing.  I can confirm that the GitHub Actions VSCode plugin successfully located the new name and since it is the same code base, it just work out the box.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -247,11 +247,10 @@ jobs:
 
       - name: Publish package distributions to GitHub Releases
         id: github-release
-
         # NOTE: DO NOT wrap the conditional in ${{ }} as it will always evaluate to true.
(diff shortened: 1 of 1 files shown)


codejedi365:
No longer blocked as we can verify that the deprecated message existed for the v9.9.0 release
Screen Shot

PR #1039, opened 2024-09-28
ci: fix release on moved main

codejedi365:
## Purpose

- Resolves release when workflows are backed up and might release on a new HEAD of main not tested
- Updated the example docs to prevent others having this issue

## Rationale

This happened for the release of `v9.9.0`.  I merged 3 PRs within a short time and since the tests were still running I thought I prevented the actual release commit pipeline from running but then the previous pipeline released the new code anyway because it pulled from the branch when the branch had moved during the workflow execution.

## How did you test?

Unfortunately I won't be able to test this until it happens again.  This should fix any issues as it is what I would do manually.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -193,10 +193,18 @@ jobs:
       contents: write
 
     steps:
-      - uses: actions/checkout@v4
+      # Note: we need to checkout the repository at the workflow sha in case during the workflow
+      # the branch was updated. To keep PSR working with the configured release branches,
+      # we force a checkout of the desired release branch but at the workflow sha HEAD.
+      - name: Setup | Checkout Repository at workflow sha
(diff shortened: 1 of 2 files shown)


codejedi365:
Validated on with 2 PR resolutions, this PR & PR #1040. As you can see in the [run](https://github.com/python-semantic-release/python-semantic-release/actions/runs/11085720277/job/30802641943), the correct checkout was the head of the PR after rebased onto main `14f04dff` however at the time `origin/master` was now at `adb1f7c6`. Therefore it properly evaluated and looked to release on the actual commit that triggered the workflow.  This fixed the issue we had with `v9.9.0` where the pipeline for #1033, with workflow sha of `d46596a9` released v9.9.0 based on `origin/master` being at `6a5d35d0`.

PR #1054, opened 2024-10-09
fix: Validate branch config regex matches and allow plain wildcards

### LINKED ISSUE #931: branches.match value "*" causes unintentional error
## The problem

If one sets the matches value of semantic_release.branches.main.match property to "*", the --version command fails with an confusing error message.
```toml
[semantic_release.branches.main]
match = "*"
prerelease_token = "rc"
prerelease = false
```
causes
```
semantic-release -vv version
[14:51:17] DEBUG    [semantic_release.cli.commands.main] DEBUG main.main: logging level set to: DEBUG  main.py:105
           DEBUG    [git.util] DEBUG util.is_cygwin_git: Failed checking if running in CYGWIN due to:  util.py:469
                    FileNotFoundError(2, 'No such file or directory')
           DEBUG    [semantic_release.cli.commands.main] DEBUG main.main: global cli options:          main.py:123
                    GlobalCommandLineOptions(noop=False, verbosity=2, config_file='pyproject.toml',
                    strict=False)
           INFO     [semantic_release.cli.util] INFO util.load_raw_config_file: Loading configuration   util.py:78
                    from pyproject.toml
           DEBUG    [semantic_release.cli.util] DEBUG util.load_raw_config_file: Trying to parse        util.py:81
                    configuration pyproject.toml in TOML format

Traceback (most recent call last):
  File "/Users//.pyenv/versions/3.10.11/bin/semantic-release", line 8, in <module>
    sys.exit(main())
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/click/core.py", line 1685, in invoke
    super().invoke(ctx)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/semantic_release/cli/commands/main.py", line 148, in main
    runtime = RuntimeContext.from_raw_config(
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/semantic_release/cli/config.py", line 306, in from_raw_config
    branch_config = cls.select_branch_options(raw.branches, active_branch)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/site-packages/semantic_release/cli/config.py", line 263, in select_branch_options
    if re.match(options.match, active_branch):
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/re.py", line 190, in match
    return _compile(pattern, flags).match(string)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/re.py", line 303, in _compile
    p = sre_compile.compile(pattern, flags)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/sre_compile.py", line 788, in compile
    p = sre_parse.parse(p, flags)
  File "/Users/julianfrenzel/.pyenv/versions/3.10.11/lib/python3.10/sre_parse.py", line 955, in parse
    p = _parse_sub(source, state, flags & SRE_FLAG_VERBOSE, 0)
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/sre_parse.py", line 444, in _parse_sub
    itemsappend(_parse(source, state, verbose, nested + 1,
  File "/Users//.pyenv/versions/3.10.11/lib/python3.10/sre_parse.py", line 669, in _parse
    raise source.error("nothing to repeat",
re.error: nothing to repeat at position 0
```

this error message is related to the * operator in the re package https://docs.python.org/3/library/re.html. The issue here is that * tries to repeat from the preceding regular expression, which is not given in this case.
## Expected behavior

A brief validation of the matches property would be great.

## Environment

Please state which OS you are using and provide the output of the following commands:

```shell
python --version 
3.10.11
pip --version
23.0.1
semantic-release --version
semantic-release, version 9.7.3
```

Please also indicate which Python build tool(s) you are using (e.g. `pip`, `build`,
`poetry`, etc.), including the version number too.
poetry v1.6.1

## Configuration

```
  72   │ [tool.semantic_release]
  73   │ version_toml = [
  74   │     "pyproject.toml:tool.poetry.version",
  75   │ ]
  76   │ version_variables = [
  77   │     "src/package_name/__init__.py:__version__",
  78   │ ]
  79   │
  80   │ [tool.semantic_release.branches.main]
  81   │ # semantic-releases makes a dry-run, so the branch name does not matter
  82   │ match = "*"
```

qthequartermasterman:
## Purpose

Resolves #931 

No regex validation is currently done in branch configs, causing an obscure error from the `re` library when the regex fails to compile. This is not obvious to the user what the error is, nor does it give any indication of how to fix it.

This is especially a problem when `match="*"`, which a user would expect to match all branches, but fails to match anything and throws this error because the plain wildcard is not a valid regex.

## Rationale

Pydantic field validation is often the correct place to validate inputs because it keeps logic simpler in other places. 

Validating a regex is easiest by trying to compile it and catching any exceptions, hence that check.

The plain wild card `"*"` is common enough an idiom where although it is not a valid regex, its intent is clear, and supporting it is simple, by adapting it into the correct regex `.*`. Additionally, in #931, some appetite for this change is expressed.

## How did you test?

I added two unit tests for these two validations. One where `*` is turned into `.*`, and one on a series of invalid regexes (other than the plain wildcard).

## How to Verify

This can be verified by modifying a branch configuration in `pyproject.toml` to `match=*` or to an invalid regex.

### DIFF
--- a/semantic_release/cli/config.py
+++ b/semantic_release/cli/config.py
@@ -143,6 +143,19 @@ class BranchConfig(BaseModel):
     prerelease_token: str = "rc"  # noqa: S105
     prerelease: bool = False
 
+    @field_validator("match", mode="after")
+    @classmethod
+    def validate_match(cls, match: str) -> str:
+        # Allow the special case of a plain wildcard although it's not a valid regex
+        if match == "*":
+            return ".*"
+
+        try:
+            re.compile(match)
+        except re.error as err:
+            raise ValueError(f"Invalid regex {match!r}") from err
+        return match
+
 
 class RemoteConfig(BaseModel):
     name: str = "origin"
--- a/tests/unit/semantic_release/cli/test_config.py
+++ b/tests/unit/semantic_release/cli/test_config.py
@@ -10,6 +10,7 @@ from pydantic import RootModel, ValidationError
 
 import semantic_release
 from semantic_release.cli.config import (
+    BranchConfig,
     GlobalCommandLineOptions,
     HvcsClient,
     RawConfig,
@@ -277,4 +278,28 @@ def test_load_invalid_custom_parser(
         RuntimeContext.from_raw_config(
             RawConfig.model_validate(load_raw_config_file(example_pyproject_toml)),
             global_cli_options=GlobalCommandLineOptions(),
+        )
+
+
+def test_branch_config_with_plain_wildcard():
+    branch_config = BranchConfig(
+        match="*",
+    )
+    assert branch_config.match == ".*"
+
+
+@pytest.mark.parametrize(
+    "invalid_regex",
+    [
+        "*abc",
+        "[a-z",
+        "(.+",
+        "{2,3}",
+        "a{3,2}",
+    ],
+)
+def test_branch_config_with_invalid_regex(invalid_regex: str):
+    with pytest.raises(ValidationError):
+        BranchConfig(
+            match=invalid_regex,
         )


codejedi365 (approved):
Great work, thanks for taking this on and completing the implementation!

PR #1098, opened 2024-11-15
Implement $CI_JOB_TOKEN support in releases

### LINKED ISSUE #977: Support CI_JOB_TOKEN now that Gitlab added the option for it to push to its own repository which will be released in 17.2 (July 18th)
## Description

Currently, semantic-release expects an Access Token to be used with Gitlab because the CI_JOB_TOKEN would not allow pushing to the repository. Now that Gitlab has enabled that, it would nice to leverage the CI_JOB_TOKEN instead of having to manage expiring tokens per project or group.

[Allow CI_JOB_TOKEN to push to its own repository #389060](https://gitlab.com/gitlab-org/gitlab/-/issues/389060)

## Use cases

* Every project requires an access token with write_repository which creates a new bot user every time
* Group access tokens can be used to limit the number of bots mentioned above, but then security takes a hit
* CI_JOB_TOKEN is better scoped and short-lived
* No expiry of tokens as Gitlab limits access tokens to 1 year now 

## Possible implementation

python-semantic-release/semantic-release/hvcs/gitlab.py L76 reads:
``` python
self._client = gitlab.Gitlab(self.hvcs_domain.url, private_token=self.token)
```

CI_JOB_TOKEN requires that the client be created as such:

``` python
self._client = gitlab.Gitlab(self.hvcs_domain.url, job_token=self.token)
```

Maybe we could add an extra parameter or check if CI_JOB_TOKEN exists when the environment variable GITLAB_TOKEN is not populated in order to determine which of the two syntax to use.

jetocotoje:
## Purpose

Implement support for the `CI_JOB_TOKEN` variable in Gitlab CI.

`CI_JOB_TOKEN`s can now be used to push to repositories, and using them is both safer and easier.

## Rationale

The `CI_JOB_TOKEN` should only complement the usage of `GITLAB_TOKEN` and respect its' value when used. When the values of `GITLAB_TOKEN` and `CI_JOB_TOKEN` match, or the remote token env name is set to `CI_JOB_TOKEN`, it is considered to be a job token. When `GITLAB_TOKEN` is not specified, the `CI_JOB_TOKEN`'s value is used.

## How did you test?

I added some parameters to the existing GitLab unit test that checks the usage of the environment variable. I also created an e2e test case that checks various combinations of personal token and job token usage in requests to the GitLab API.

## How to Verify

1. Enable usage of job tokens for pushing in the Gitlab repository configuration
2. Set the `tool.semantic_release.remote.token` to `{ env = "CI_JOB_TOKEN" }` or assign `export GITLAB_TOKEN = $CI_JOB_TOKEN` in a GitLab CI configuration.
3. Trigger a release/version job in the pipeline

### DIFF
--- a/docs/configuration/configuration.rst
+++ b/docs/configuration/configuration.rst
@@ -1125,6 +1125,14 @@ default token value will be for each remote type.
 **Default:** ``{ env = "<envvar name>" }``, where ``<envvar name>`` depends on
 :ref:`remote.type <config-remote-type>` as indicated above.
 
+A special case is the GitLab CI environment variable ``"CI_JOB_TOKEN"``. If this variable is set,
+it will be used as the job token for the GitLab API. This is useful for accessing the Gitlab Releases
+API in the CI environment.
+
+.. warning::
+    The value of ``"GITLAB_TOKEN"`` takes precedence over ``"CI_JOB_TOKEN"``. If both are set to the
+    same value, it is assumed to be a job token, not a personal access token.
+
 ----
 
 .. _config-remote-type:
--- a/src/semantic_release/hvcs/gitlab.py
+++ b/src/semantic_release/hvcs/gitlab.py
@@ -47,7 +47,6 @@ class Gitlab(RemoteHvcsBase):
         **_kwargs: Any,
     ) -> None:
         super().__init__(remote_url)
-        self.token = token
         self.project_namespace = f"{self.owner}/{self.repo_name}"
         self._project: GitLabProject | None = None
 
@@ -67,18 +66,23 @@ class Gitlab(RemoteHvcsBase):
                 path=str(PurePosixPath(domain_url.path or "/")),
             ).url.rstrip("/")
         )
-        private_token, job_token = token, None
+
+        private_token = token
         job_token = os.getenv("CI_JOB_TOKEN")
 
+        self.token = private_token
         if job_token:
             if job_token == private_token or not private_token:
                 # Disable private_token if it's actually the CI_JOB_TOKEN
                 private_token = None
+                self.token = job_token
             else:
                 # Private token should be prioritized over CI_JOB_TOKEN
                 job_token = None
 
-        self._client = gitlab.Gitlab(self.hvcs_domain.url, private_token=private_token, job_token=job_token)
+        self._client = gitlab.Gitlab(
+            self.hvcs_domain.url, private_token=private_token, job_token=job_token
+        )
         self._api_url = parse_url(self._client.api_url)
 
     @property
--- a/src/semantic_release/hvcs/gitlab.py
+++ b/src/semantic_release/hvcs/gitlab.py
@@ -67,8 +67,18 @@ class Gitlab(RemoteHvcsBase):
                 path=str(PurePosixPath(domain_url.path or "/")),
             ).url.rstrip("/")
         )
+        private_token, job_token = token, None
+        job_token = os.getenv("CI_JOB_TOKEN")
 
-        self._client = gitlab.Gitlab(self.hvcs_domain.url, private_token=self.token)
+        if job_token:
+            if job_token == private_token or not private_token:
+                # Disable private_token if it's actually the CI_JOB_TOKEN
+                private_token = None
+            else:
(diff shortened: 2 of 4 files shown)


codejedi365:
@KundaPanda, this is a great contribution as you also included tests and documentation, Thank you! I just want to add a few things and we will get this merged in and released.

jbnjohnathan:
Any update on this? From the comments it seemed pretty ready for merge

codejedi365:
I'll take a look at it again this weekend. It got majorly complicated because it was a bit premature to add this functionality without handling all the variations of GitLab installations. 

After digging through GitLab's issue queue, I found that the CI_JOB_TOKEN support was only provided behind a feature flag that was turned off by default. Then I found that the official GitLab (gitlab.com) did not activate this feature flag because of a secondary issue ticket of infinite looping pipelines. This added undo complexity to attempting to determine what error states could happen. 

The goal was to support GitLab proper, GitLab on-prem of multiple versions (no feature flag, feature flag disabled, and feature flag enabled). Each circumstance would provide some level of error message to the user of why it's not working otherwise I get a lot of issue tickets about why it isn't working.

codejedi365:
> Hi @codejedi365 thanks for trying to implement this properly and support all deployment variants. If you need any testing, we could test this on our gitlab instance or even provide you with a token/user to test it yourself.

Wow, why thank you. I'll let you know if I need a testing environment. I do have a separate GitLab instance to test against when I get to that part but I appreciate the gesture.

> Can you refer to those issues/feature flags? In general, CI_JOB_TOKEN should be available in all environments afaik. Is it a specific permission which is missing?

The job token has been available for a long time, but the problem has been permissions. It was restricted from writing to the repository. It has specific permissions hardcoded by GitLab. This is unlike GitHub which enabled users to define the permissions of the token per workflow or job. In GitLab v17.2 (see #977), it was provided as a feature flag that admins could enable but it was disabled by default. GitLab.com does not expose feature flag enablement to users and so it is disabled.

Original: https://gitlab.com/gitlab-org/gitlab/-/issues/389060
Roll-out plan: https://gitlab.com/gitlab-org/gitlab/-/issues/468320
Which then Scott Hampton mentions, it can't roll out until https://gitlab.com/gitlab-org/gitlab/-/issues/475705

----

Oddly enough even as this permission becomes available, previously you could write to the releases endpoint via the API and it will create the associated tag at a given commit sha. The problem is that PSR was not written to support this by the original maintainers.  In this circumstance, you also must point the release at an existing commit in the remote which would require `--no-commit` to be set on PSR.  Consequently how do we handle an updating changelog because we can't commit it with each release.  I bring this up because GitLab has this issue but so does GitHub if you choose to add additional protections to your default branch. I feel we are in a no-win solution currently for better security and automated releases.  Not to mention the importance of limiting subsequent pipelines via GitLab's workflow rules and GitHub's action event triggers which I get too many issues reported about. I digress.

benedikt-bartscher:
It seems like there is some movement on gitlab's side. They are currently rolling out the feature flag in production... Let's hope everything goes well and we can finally have secure granular permissions without configuration overhead :pray:

benedikt-bartscher:
Gitlab has rolled out the feature on almost all deployments - GA is planned for the next feature release 18.4 (https://gitlab.com/gitlab-org/gitlab/-/issues/468320#note_2679031928). 18.4 should be released in about 2 weeks.

Hopefully we can finally get this one in. If you need any help/testing, feel free to ping me

codejedi365:
Thanks for the update! I should be able to work on this after finishing the monorepo parser PR.

benedikt-bartscher:
Hi @codejedi365, if you need any help on this please let me know

eran-gil2:
Hey, is there any update on this? To scale our semver workflow across the organization we need this feature to allow a unified job definition across projects while also avoiding creation of a "too powerful" token.
Do you have an ETA for the completion? @codejedi365

benedikt-bartscher:
I am still highly interested in this

benedikt-bartscher:
@codejedi365 are you still interested in getting this in? I would put some effort in development and testing with a big gitlab setup where a lot of repos use python-semantic-release.

Sorry for spamming here, but i'd like to prevent putting work in this if there is no chance to get it in.

Issue #1136, opened 2025-01-10
`changelog.exclude_commit_patterns` does not filter release commits


HADB:
### Description

I have a custom `commit_message` and I have set `changelog.exclude_commit_patterns` to ignore it in changelog, but it will still be generated in `CHANGELOG.md` (except for the last release):

Here is my config：

```toml
[tool.semantic_release]
    commit_message = "chore(release): v{version}\n\nAutomatically generated by python-semantic-release"
    commit_parser = "angular"
    tag_format = "v{version}"
[tool.semantic_release.changelog]
    exclude_commit_patterns = ['''chore: release|chore\(release\): .+''']
[tool.semantic_release.commit_parser_options]
    allowed_tags = [
        "build",
        "chore",
        "ci",
        "docs",
        "feat",
        "fix",
        "perf",
        "refactor",
        "style",
        "test",
    ]
    minor_tags = ["feat"]
    patch_tags = ["chore", "ci", "docs", "fix", "perf", "refactor"]
```

Here is the generated `CHANGELOG.md`
```md
# CHANGELOG


## v0.6.10 (2025-01-10)

### Chores

- Update project.toml
  ([`8660f6a`](https://github.com/font2svg/font2svg-api/commit/8660f6aa2d8417427e170619a72d6a34baa0a46b))


## v0.6.9 (2025-01-10)

### Chores

- Update pyproject.toml
  ([`5dacf17`](https://github.com/font2svg/font2svg-api/commit/5dacf17a3e2a5eb227161f51cc8d3e086cdcc08f))

- **release**: V0.6.9
  ([`a968289`](https://github.com/font2svg/font2svg-api/commit/a9682898f032eb81c999d31d61b20504279be013))

...
``` 

## Version

`9.15.2`

## GitHub Action logs

https://github.com/font2svg/font2svg-api/actions/runs/12704060826/job/35412829762

codejedi365:
This is happening because a chore commit is slated to cause a patch change in your configuration. The exclude pattern is ignored because your version commit is marked as chore and in one of the recent updates, I made it so that PSR would not hide a commit that caused a version bump. 

It's a bit abnormal for a chore commit to cause a version bump. PSR would normally ignore a version commit it would make but since I made the change to not hide commits that bump versions this was no longer the case in your configuration. Not quite my intention so I'll fix that.

codejedi365:
[...]

PR #1139, opened 2025-01-12
fix(changelog): fixes PSR release commit exclusions for customized commit messages

codejedi365:
## Purpose

- Fix PSR to properly ignore PSR release commit messages when customized

## Rationale

I came across this bug when looking into #1136. Although #1136 has a faulty configuration, it ends up I found that PSR was not safely escaping the custom release message prior to making it a regular expression. In this case we could of had a weird bug where the regular expression started to throw an error but most have been silent bugs.  The more important situation to cover is when people have modified the commit message to strictly adhere to conventional commits even for their release commits. This can potential add characters for the scope that otherwise have special meanings in regular expressions which then leads to their commit not being ignored.  This PR prevents that from happening.  

Secondly, this PR also adds some regular expression validation into the configuration validation steps. This should help identify problems early for people.

## How did you test?

Added unit tests to handle the validation verification both invalid and valid regular expressions to include different numbers of patterns.  In the E2E tests, I added one test that rebuilds the last two releases with setting a custom commit message format.  It is then validated that the changelog is the same as the original and does not have the release commit included in the changelog for the previous version.  When reviewing the test verification steps below the the 2 init mode changelogs fail because the custom commit message failed to be excluded.

## How to Verify

```sh
# 1. Check out this PR
git fetch upstream/pull/1139:pr-1139
git checkout --detach pr-1139~3

# 2. Run pytest & observe the failures
pytest -vv -k test_changelog_config_with_
pytest -vv -k test_version_changelog_content_custom_commit_message_excluded_automatically
# Note: in the end-to-end tests, there are 2 passing but that is expected because of
# how Update mode works. It doesn't touch the previous version, which is where a
# release commit will appear.

# 3. Update to the head of the PR
git merge pr-1139

# 4. Run pytest again and all tests are resolved
pytest -vv -k test_changelog_config_with_
pytest -vv -k test_version_changelog_content_custom_commit_message_excluded_automatically
```


codejedi365:
### :tada: This PR has been published as part of Version 9.16.0 :tada:

You can find more information about this release on the [GitHub Releases](https://github.com/python-semantic-release/python-semantic-release/releases/tag/v9.16.0) page.

PR #1142, opened 2025-01-12
fix(parser-custom): handle relative parent directory paths to module file better

codejedi365:
## Purpose

- Avoids any issues with dynamic importing given a file name with parent path traversals

## Rationale

The dynamic import originally would just replace "/" with "." to make the import module name more pythonic, however this would be problematic in monorepos which would use "../../misc/commit_parser.py" to locate the parser and so the resulting `sys.modules` entry would have numerous periods (.) as a prefix. This removes that possibility. Still always an issue if the imported module name matches an existing module but the likelihood is low.

## How did you test?

Used my codejedi365/psr-monorepo-poweralpha repository to test a custom monorepo parser and release

### DIFF
--- a/src/semantic_release/helpers.py
+++ b/src/semantic_release/helpers.py
@@ -69,7 +69,6 @@ def dynamic_import(import_path: str) -> Any:
     Dynamically import an object from a conventionally formatted "module:attribute"
     string
     """
-    log.debug("Trying to import %s", import_path)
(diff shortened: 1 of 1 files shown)


codejedi365:
### :tada: This PR has been published as part of Version 9.16.1 :tada:

You can find more information about this release on the [GitHub Releases](https://github.com/python-semantic-release/python-semantic-release/releases/tag/v9.16.1) page.

Issue #1150, opened 2025-01-22
Request support git "insteadOf" url aliases


larsks:
### Description

With a configuration like this in my global git config:

```
[url "ssh://git@github.com/larsks/"]
	insteadof = me:
```

I can clone a repository like this:

```
git clone me:myproject
```

Which ends up looking like this:

```
$ git config --get remote.origin.url
me:myproject
```

Which ultimately causes psr to fail like this:

```
$ psr version
::ERROR:: Bad url: 'me:myproject'
```

### Expected behavior

PSR would not have thrown an error because it would have been able to interpret my git configuration and extracted the remote URL just as git normally can.

If you use `git remote get-url <remote>` command, it will return a fully expanded url via the `insteadOf` setting.

```
$ git remote get-url origin
ssh://git@github.com/larsks/myproject
```

### Actual behavior

PSR errored out with a failure message of a bad remote URL because it retrieved the remote url setting which in raw form is an invalid URL rather than using the git interpreter which supports the `insteadOf` alias attribute and would of returned a valid remote URL value.

### Environment

```
semantic-release, version 9.16.1
```

codejedi365:
I don't appreciate the rude tone as I do this on my own time for $0.

larsks:
Look, I wasn't trying to be rude. I was just opening an issue. I'm sorry to bother you; I'll work on fixing it myself and submit a PR at some point.

larsks:
I've re-read the original post, and I'm not sure what you've identified as "rude". Was it this?

> psr fell over because it couldn't parse a valid remote specification.

Maybe the description is a bit terse, but I think it accurately captures the behavior. I apologize if it came across as rude. For what it's worth, I *also* work on software for $0 in my spare time, so we have in common, at least.

If other folks want to tackle this, note that the underlying problem is that PSR relies on GitPython, which has been in "maintenance mode" for at least eight years. The [most relevant issue](https://github.com/gitpython-developers/GitPython/issues/571) was opened backed in 2017, and has remained untouched since then.

The correct path forward is probably to fix this in GitPython, but note that even if it gets fixed there, PSR will need to update its dependencies, which currently pins GitPython to `~= 3.0`, whereas GitPython has moved on to 3.1.xx.

(The alternative solution would be to abandon GitPython and use the `git` cli, which would resolve all the compatibility issues but would require `git` to be installed; the current solution avoids that external dependencies, so maybe it's a wash.)

larsks:
[...]

PR #1151, opened 2025-01-22
fix: Deal with remote urls aliased using insteadOf directives

larsks:
## Purpose

- Add support for git insteadOf configuration for repository origin
- Resolves: #1150

## Rationale

Previously, python-semantic-release would attempt to use the raw url from
the repository configuration (`me:myproject` in the above example), and
would fail to parse it because it's not actually a url. With this change,
psr instead calls out to `git remote get-url`, which fully expands any aliases via git.

[1]: https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf

## How I tested

Created a unit test that will modify a git configuration with a remote origin defined with an alias and the associated alias defined separately.  The runtime context is loaded and then validated through the GitHub VCS client.

## How to Verify

```sh
git fetch pull/1151/head pr-1151
git checkout pr-1151~1 --detach

# validate the added test will replicate the failure on the original code
pytest -vv -k test_git_remote_url_w_insteadof_alias

# Checkout the latest
git checkout pr-1151

# Run the tests again and see the failure is resolved
pytest -vv -k test_git_remote_url_w_insteadof_alias
```

### DIFF
--- a/src/semantic_release/cli/config.py
+++ b/src/semantic_release/cli/config.py
@@ -602,7 +602,12 @@ class RuntimeContext:
         # Retrieve details from repository
         with Repo(str(raw.repo_dir)) as git_repo:
             try:
-                remote_url = raw.remote.url or git_repo.remote(raw.remote.name).url
+                # Get the remote url by calling out to `git remote get-url`. This returns
+                # the expanded url, taking into account any insteadOf directives
+                # in the git configuration.
+                remote_url = raw.remote.url or git_repo.git.remote(
+                    "get-url", raw.remote.name
+                )
                 active_branch = git_repo.active_branch.name
             except ValueError as err:
                 raise MissingGitRemote(
--- a/src/semantic_release/hvcs/_base.py
+++ b/src/semantic_release/hvcs/_base.py
@@ -30,7 +30,7 @@ class HvcsBase(metaclass=ABCMeta):
     """
 
     def __init__(self, remote_url: str, *args: Any, **kwargs: Any) -> None:
-        self._remote_url = remote_url
+        self._remote_url = remote_url if parse_git_url(remote_url) else ""
         self._name: str | None = None
         self._owner: str | None = None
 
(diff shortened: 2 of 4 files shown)


codejedi365:
@larsks, thanks for the PR resolution, I'll try to have it into this weeks release. If there is a unit test we can add maybe to `test_config` that would be great. I'm thinking we probably don't need the `or` statement and should just default to the get url?

larsks:
> If there is a unit test we can add maybe to test_config that would be great. 

Sure, I'll try to take a look at that this evening.

> I'm thinking we probably don't need the or statement and should just default to the get url?

You mean here?

```
remote_url = raw.remote.url or git_repo.remote(raw.remote.name).url
```

I suspect you're correct; I wasn't sure where/how `raw.remote.url` was set. I'll make this change as well.

codejedi365:
> I suspect you're correct; I wasn't sure where/how `raw.remote.url` was set. I'll make this change as well.

Oh wait, disregard. The `or` is correct. As the raw is the actual configuration from the user. They can pass the URL in the configuration.

Let me know if you need me to build the test. You are going to want to use on of our default repo fixtures. The easiest and fastest is repo_initial_commit.

larsks:
I've put together a test, but it's the first test that manipulates the git configuration so there wasn't a model to follow. Let me know what you think.

codejedi365:
> I've put together a test, but it's the first test that manipulates the git configuration so there wasn't a model to follow. Let me know what you think.

It was a good skeleton, Thanks! Made some adjustments and expanded the tests to run across all the VCS types as they all implement `remote_url()`.

codejedi365:
### :tada: This PR has been published as part of Version 9.17.0 :tada:

You can find more information about this release on the [GitHub Releases](https://github.com/python-semantic-release/python-semantic-release/releases/tag/v9.17.0) page.

Issue #1169, opened 2025-02-03
commands are failing for self hosted gitea release hosted via http


ste1nstone:
### Description
Hi I think I encountered the same problem like described in the bug  #1074 today.
I tried to run semantic release on an action with my self hosted gitea instance that is hosted via http.

![Image](https://github.com/user-attachments/assets/ee51d550-be3d-40de-9960-e9cf37b72aab)

Here is my remote config 
```
[semantic_release.remote]
name = "origin"
domain = "http://gitea:3000"
type = "gitea"
insecure = true
```

what I noticed during my investigations was that gitea sets the same env variables like github. So the following environment variables are set in the pipeline: 
- GITHUB_SERVER_URL
- GITHUB_API_URL

For testing I tried to unset the two env vars before executing semantic release it worked.
```
unset GITHUB_SERVER_URL
unset GITHUB_API_URL
```

I assumtion is that semantic release somehow does a remote auto config based on the env variables.
But I do not understand the code enough to validate that theory.

It can also be reproduced without a cicd pipeline
Just set the GITHUB_SERVER_URL env variable with a http schema and run semantic-release --version

```
$env:GITHUB_SERVER_URL = "http://gitea:3000"
semantic-release --version
```

This already leads to the same issue on my local windows pc

Hope this info helps someone with fixing the issue.

BR
Philipp

codejedi365:
@ste1nstone, thanks for the report. I think I understand what is happening now as the previous issue the author never responded. It's loading the default config object before reading your configuration and then reads your configuration but it throws an error before it reads your configuration. Not quite what was expected.

codejedi365:
@ste1nstone, would you test out the following pre-release to see if it resolves the issue in your environment?

https://github.com/codejedi365/python-semantic-release/releases/tag/v9.18.1-alpha.1

ste1nstone:
@codejedi365 with the new version it now also works without having to unset the two variables first.
Thanks for the fix and the cool project

codejedi365:
@ste1nstone, you're welcome!

codejedi365:
### :tada: This issue has been resolved in Version 9.18.1 :tada:

You can find more information about this release on the [GitHub Releases](https://github.com/python-semantic-release/python-semantic-release/releases/tag/v9.18.1) page.

Issue #1074, opened 2024-10-31
An error occurred in the workflow


michaelbennn:
### Description

I am going to use python-semantic-release to perform CICD operations in gitea action, but it reports an error and it keeps this error

![image](https://github.com/user-attachments/assets/c259ef3c-0d2c-474e-a722-2e632ea5c9f4)

this is workflow.
![image](https://github.com/user-attachments/assets/43cb031d-55ca-41ba-ae5a-8fcf70c65fa4)

### Environment

-  **Operating System (w/ version):** 
-  **Python version:** 
-  **Pip version:** 
-  **Semantic-release version:** 
-  **Build tool (w/ version):** 

<details>
<summary><code>pip freeze</code></summary>

```log

```

</details>

<br>

<details>
<summary><code>git log --oneline --decorate --graph --all -n 50</code></summary>

```log

```

</details>

### Configuration

<details>
<summary>Semantic Release Configuration</summary>

```toml
```

</details>
<br>

<details>
<summary>Build System Configuration</summary>

```toml
```

</details>
<br>

<details>
<summary>GitHub Actions Job Definition</summary>

```yaml

```

</details>

### Execution Log

<details>
<summary><code>semantic-release -vv <i>command</i></code></summary>

```log

```

</details>

codejedi365:
@michaelbennn, that error is trying to let you know that there is an error in your configuration specifically in the remote configuration section. 

Can you fill out the bug template with your configuration at least? The boilerplate is supposed to help me determine the issue and provide a solution. 

The specific error is detecting that you have configured a HTTP URL of some kind and have not set the `remote.insecure` flag to true to enable it.

codejedi365:
Also don't point at the master branch, always point at a specific tag when defining an action. If I introduced breaking changes or even just had master branch in a broken state, you would have a broken pipeline.

codejedi365:
### :tada: This issue has been resolved in Version 9.18.1 :tada:

You can find more information about this release on the [GitHub Releases](https://github.com/python-semantic-release/python-semantic-release/releases/tag/v9.18.1) page.

PR #1173, opened 2025-02-07
fix(config): refactors default token resolution to prevent insecure URL error

codejedi365:
## Purpose

- Resolves: #1169
- Resolves: #1074

## Rationale

Due to the previous default token loading, PSR configuration load would initate the default RemoteConfig. During validation of the remote config, the HVCS object was initialized and upon initialization the URL is evaluated for insecure protocol causing errors when the URL was pulled from the env.

This change removes the need for initialization which means the URL validation will not occur pre-maturely

## How did you test?

Allowed user to test in their environment and it worked for them. No automated test was created.

### DIFF
--- a/src/semantic_release/cli/config.py
+++ b/src/semantic_release/cli/config.py
@@ -54,7 +54,6 @@ from semantic_release.errors import (
     ParserLoadError,
 )
 from semantic_release.helpers import dynamic_import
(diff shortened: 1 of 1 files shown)


codejedi365:
### :tada: This PR has been published as part of Version 9.18.1 :tada:

You can find more information about this release on the [GitHub Releases](https://github.com/python-semantic-release/python-semantic-release/releases/tag/v9.18.1) page.

Issue #1149, opened 2025-01-20
Question: How to handle branch protections with PSR?


ichoosetoaccept:
## Question
I'm trying to understand if it's possible to use python-semantic-release in a way that ONLY creates GitHub releases, version tags, and release notes WITHOUT requiring commits/pushes to the main branch.

My use case: I have a protected main branch with required status checks ([Codacy](https://www.codacy.com/)). I want semantic-release to run AFTER a PR is merged (meaning all checks have passed), but ONLY to create the release artifacts without modifying the repository.
Current behavior: The tool attempts to commit and push changes even when I think I've configured it not to, causing failures due to branch protection rules.

Is there a way to achieve this, or is committing/pushing an essential part of semantic-release's workflow that cannot be bypassed?

The original authors of the [semantic-release project](https://semantic-release.gitbook.io/semantic-release) [strongly recommend](https://semantic-release.gitbook.io/semantic-release/support/faq#making-commits-during-the-release-process-adds-significant-complexity) **against** altering project state during a release and I'm trying to heed their advice.

### Configuration
- [pyproject.toml](https://github.com/ichoosetoaccept/unifi_assist/blob/main/pyproject.toml)
- [release.yml](https://github.com/ichoosetoaccept/unifi_assist/blob/main/.github/workflows/release.yml)

## Additional context
I've tried various combinations of the GitHub Action configuration, including removing explicit commit: false and push: false settings, but python-semantic-release still attempts to push changes. My GitHub workflow is configured to run on PR merge to main (see Configuration above).

Here's a [full log](https://github.com/ichoosetoaccept/unifi_assist/actions/runs/12877252573/job/35901418211#step:4:1) with -vv active. [It shows](https://github.com/ichoosetoaccept/unifi_assist/actions/runs/12877252573/job/35901418211#step:4:2287) that version detection is working perfectly but it never creates a Github release with release notes and a version bump.

I was able to get the desired outcome [once earlier](https://github.com/ichoosetoaccept/unifi_assist/releases/tag/0.1.0) but that was **before**  I added a branch protection ruleset to protect the main branch:

- Release 0.1.0 was created on: 2025-01-20T09:58:01Z
- The ruleset was created on: 2025-01-20T12:38:33.202+01:00 (about 2.5 hours after the release)

The ruleset "Protect main" includes:
- Protection against branch deletion
- Required status checks from Codacy Static Code Analysis
- Nothing else

[...]

PR #1143, opened 2025-01-12
feat(parser-conventional-monorepo): add new conventional-commits standard parser for monorepos

codejedi365:
## Purpose

Adds a conventional commits parser designed for monorepos

## How did you test?

Added 3 new monorepo fixtures and rebuild tests for various types of monorepo workflows. Specifically:
- Trunk only development for a 2 package monorepo
- GitHub Flow development strategy for a 2 package monorepo with squash commits
- GitHub Flow development strategy for a 2 package monorepo with merge commits

PR #1216, opened 2025-03-14
feat(parser-angular): add scope filtering option to Angular commit pa…

### LINKED ISSUE #1215: Improve support for monorepo using Conventional Commits  scope
### Description
Currently, **Python Semantic Release** does not support filtering commits by **scopes (`(pkg1)`, `(pkg2)`)** in monorepos. This means that when **multiple packages exist in a single repository**, version bumps may happen incorrectly.

For example, given a monorepo with two packages: `/packages/pkg1`, `/packages/pkg2`

If a commit like:

```plaintext
feat(pkg1): Added new API feature
```

appears, **pkg2 should not get a version bump**, but currently, `semantic-release` does not support filtering commits by scope.

### Use cases
1. **Monorepos with independent package versioning**  
   - Developers working on multiple packages in a single repository often want **each package to be released separately**.
   - Example: A commit `feat(pkg1): Added logging` should bump **pkg1's version**, but **pkg2's version should remain unchanged**.

2. **Scoped commit filtering for better release control**  
   - Some teams use commit scopes (`feat(pkg1): ...`, `fix(pkg2): ...`) to **specifically indicate which package was modified**.
   - Currently, `semantic-release` **reads all commits**, making it difficult to version packages separately in monorepos.

### Possible implementation
1. **Introduce a `commit_scope_filter` option in `pyproject.toml`**

```toml
[semantic_release.commit_parser_options]
minor_tags = ["feat"]
commit_scope_filter = "pkg1"
```
- If set, `semantic-release` **only considers commits with the matching scope**.
- If not set, `semantic-release` behaves as it does today (reading all commits).

2. **Modify commit parsing to respect the scope filter**  
- When calculating the next version, **ignore commits that do not match the specified `commit_scope_filter`**.

3. **Extend `semantic-release changelog` to support scope filtering**

```
semantic-release changelog --scope=pkg1
```

### Alternative solutions
- Using **separate repositories** for each package (not ideal for monorepos).
- Manually filtering commits before running `semantic-release` per package.
- Running `semantic-release` in isolated environments per package and limiting `git log` manually.

This feature would greatly improve **monorepo support** in `semantic-release` by ensuring that **only relevant commits impact versioning for each package**.

Best,
Asaf

asaf:
## Fix: Ensure Scope Filtering is Enforced Correctly

Resolves #1215

---

## Purpose
This PR ensures that **scope filtering logic is correctly applied** when `allowed_scopes` is configured in `ConventionalCommitParserOptions`. Previously, commits without a scope were incorrectly bypassing the filtering, and commits with no allowed scopes defined were still being evaluated unnecessarily.

---

## Rationale
1. If `allowed_scopes` **is not defined**, the filtering logic should be skipped entirely to maintain the existing behavior.
2. If `allowed_scopes` **is defined**, the commit **must** have a valid scope that is explicitly allowed.
3. If a commit **has no scope**, it **should not be allowed** when `allowed_scopes` is configured.

This change ensures a strict enforcement of scope filtering while keeping backwards compatibility when `allowed_scopes` is not set.

---

## How did you test?
- Added unit tests to verify:
  - Commits **with allowed scopes** pass the filter.
  - Commits **with disallowed scopes** are correctly rejected.
  - Commits **without a scope** are **rejected** when `allowed_scopes` is enabled.
  - Commits **without an `allowed_scopes` definition** behave as they currently do.

- Ran the test suite and verified that existing functionality remains unchanged for users who do not configure scope filtering.

---

## How to Verify
1. Run `pytest` to ensure all tests pass.
2. Use the `ConventionalCommitParser` with different configurations:
   - No `allowed_scopes`: Behavior remains unchanged.
   - `allowed_scopes` defined: Only commits with matching scopes should be allowed.
   - Commit with no scope: Should be rejected when `allowed_scopes` is defined.

---

## PR Completion Checklist

- [x] Reviewed & followed the [Contributor Guidelines](https://python-semantic-release.readthedocs.io/en/latest/contributing.html)
- [x] Commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard
- [x] Appropriate End-to-End tests added/updated
- [x] Appropriate Documentation added/updated and syntax validated for sphinx build (see Contributor Guidelines)

### DIFF
--- a/src/semantic_release/commit_parser/angular.py
+++ b/src/semantic_release/commit_parser/angular.py
@@ -94,6 +94,9 @@ class AngularParserOptions(ParserOptions):
     one of these prefixes, it will not be considered a valid commit message.
     """
 
+    allowed_scopes: Tuple[str, ...] = ()
+    """If set, only commits with a matching scope will be considered."""
+
     default_bump_level: LevelBump = LevelBump.NO_RELEASE
     """The minimum bump level to apply to valid commit message."""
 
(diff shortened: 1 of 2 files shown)


codejedi365:
This is already solved in #1149, but thank you for putting in the time to provide a solution.

asaf:
Closing in favor of the monorepo parser introduced in #1143

PR #1226, opened 2025-03-31
Updated github actions example - gh-action-pypi-publish path

bartdorlandt:
Fixing the PyPI github action path.
pypa/gh-action-pypi-publish@release/v1

## Purpose
Fixes the path for the github action for using pypi publishing

## Rationale

I used the example in my pipeline and it crashed on not finding the path. 

`Missing download info for pypa/gh-action-pypi-publish@v1`

## How did you test?
I used the correct path from another working pipeline.

## How to Verify
My project is using that path.
https://github.com/bartdorlandt/convert_poetry2uv

### DIFF
--- a/docs/automatic-releases/github-actions.rst
+++ b/docs/automatic-releases/github-actions.rst
@@ -735,7 +735,7 @@ to the GitHub Release Assets as well.
               git_committer_email: "actions@users.noreply.github.com"
 
           - name: Publish | Upload package to PyPI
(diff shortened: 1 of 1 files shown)


codejedi365:
Changed with PR #1229 to recommend SHA_HASH with a comment of which version it is. This change now is obsolete but still was accomplished to be more clear that end users must make the configuration work as that is not the PSR team's job.

PR #1335, opened 2025-09-21
Run tests on Python 3.14 release candidate 3

cclauss:
https://devguide.python.org/versions

## How to Verify

---

## PR Completion Checklist

- [ ] Reviewed & followed the [Contributor Guidelines](https://python-semantic-release.readthedocs.io/en/stable/contributing/contributing_guide.html)

- [ ] Changes Implemented & Validation pipeline succeeds

- [ ] Commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard

- [ ] Appropriate Unit tests added/updated

- [ ] Appropriate End-to-End tests added/updated

- [ ] Appropriate Documentation added/updated and syntax validated for sphinx build (see Contributor Guidelines)


codejedi365:
> https://devguide.python.org/versions -- Python 3.14 is scheduled to be released on Oct. 7th.

Thanks for the heads up.

> Tools that depend on `python-semantic-release` are unable to test their compatibility with Python 3.14 easily.

This is actually dependency resolution problem that is not due to python-semantic-release. As we can see in the CI logs, it's related to the pydantic official version. It fails when it downloads 2.11.9 because it has a strict dependency on `pydantic-core==2.33.3`. UV succeeds because you set it to download an alpha version which resolves the compatibility error. Most package managers (pip in this case) do not download prereleases unless you tell it to, which is the correct way to approach dependency resolution in a production environment. Officially, The dependencies of PSR are not ready for testing compatibility with Python 3.14 which in turn hinders you. 

But if Python is adhering to SemVer (which generally they do) there should be no compatibility problem with your code as it is a minor version bump. The dependencies you are using however if they hardcode or hard restrict the maximum version (like pydantic-core's py03 package) is then there might be dependencies that are unable to install on the new version. In my opinion this is not a true compatibility issue of the code but arbitrary restraints the author has placed on it (likely also a valid restriction given it's a rust module which requires specific Python version header files).

Also please also adhere to the contributing guide and PR template since I was quite confused when your PR actually doesn't fix anything.

cclauss:
Sorry.  The first commit demonstrates that a problem exists.  The second commit provides a solution that would allow projects that depend on `python-semantic-release` to easily test their software on Python 3.14 release candidate 3.

codejedi365:
Although I appreciate you providing a recommended change, I think my main point that it is not a PSR problem was misunderstood. I will not modify a production codebase to rely on an alpha release. You should open an issue on the pydantic and pydantic-core libraries to ask them when they will release their next official version. This is what happens when you test prematurely to a new release when then world is not as prepared as you are.

cclauss:
https://pydantic.dev/articles/pydantic-v2-12-release

codejedi365:
@cclauss, well that's good news. Did you try your project install test now that it is released? It should work now without the `--with` parameters.

Issue #975, opened 2024-07-07
Ability to add labels/comments to linked issues/PRs


agkphysics:
## Description

The NodeJS implementation of semantic-release has the ability to automatically tag relevant PRs and issues with specified labels with the [`@semantic-release/github`](https://github.com/semantic-release/github) plugin. It would be useful for this Python version to also be able to do so.

## Use cases

You can then easily see which issues and PRs have been released, and on which versions. Technically this is already possible for PRs by looking at the tags associated with the merge commit, but that requires extra steps.

## Possible implementation

Since there is planned functionality for linking PRs into changelogs, presumably said PRs could be passed to a step that adds labels to them, or adds a comment on them.

Issues may be slightly more difficult as they are linked via PRs.

codejedi365:
@agkphysics, I agree it is a very nice feature that I have been doing manually for this project because of how nice it is. It is an eventual goal but we have some minor stuff to solve first

codejedi365:
still valuable, no effort towards implementation has been made yet.

codejedi365:
Setting some ground work for this feature by enabling commit parsers to extract the issue numbers and PR numbers from commit messages.

codejedi365:
Under initial development and design, will require a lot more API definitions for a set of issue tracking systems. Significant amount of work to be done.

codejedi365:
Still in backlog

codejedi365:
no movement here yet, still in backlog.

codejedi365:
Still in backlog

lowlydba:
@codejedi365 I see there's a somewhat-mature, somewhat-related open PR for this still - any updates or clarity you can provide? I've found this to be a high value quality of life feature for many engineers

PR #1362, opened 2025-11-05
feat: refs for linked issues in conventional commit parser

JCHacking:
## Purpose
With conventional commit, you can also reference issues with `Refs: #1234`. This does not seem to have been considered.

## Rationale
Add that case to the regex as well.

## How to Verify
With a commit containing that reference form, when generating the changelog, it is added to linked_issues.

---

## PR Completion Checklist

- [X] Reviewed & followed the [Contributor Guidelines](https://python-semantic-release.readthedocs.io/en/stable/contributing/contributing_guide.html)

- [X] Commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard

- [x] Appropriate End-to-End tests added/updated

- [x] Appropriate Documentation added/updated and syntax validated for sphinx build (see Contributor Guidelines)


codejedi365:
> With conventional commit, you can also reference issues with `Refs: #1234`. This does not seem to have been considered.

In the [Conventional Commit Specification](https://www.conventionalcommits.org/en/v1.0.0), it only defines the authorized existence of footers, excerpt:

- _One or more footers MAY be provided one blank line after the body. Each footer MUST consist of a word token, followed by either a :<space> or <space># separator, followed by a string value (this is inspired by the [git trailer convention](https://git-scm.com/docs/git-interpret-trailers))._

- _A footer’s token MUST use - in place of whitespace characters, e.g., Acked-by (this helps differentiate the footer section from a multi-paragraph body). An exception is made for BREAKING CHANGE, which MAY also be used as a token._

- _A footer’s value MAY contain spaces and newlines, and parsing MUST terminate when the next valid footer token/separator pair is observed._

The examples used in the specification do use the `Refs:` git trailer to define issue examples, however, none of the current Team Version Control System (VCS) Servers (Github, GitLab, Gitea, BitBucket) recognize `refs` as a defined git trailer or syntax to close an issue automatically upon push or merge of that commit to the default branch.  That is where I came up with the authorized git footer list to extract issue references. In PSR's custom template, I use the `linked_issues` to add a closure reference into the CHANGELOG and I deliberately left off `ref:` because then I had a way of still adding a reference for why this commit was being made without closing it or marking it as closed in the CHANGELOG.  The `ref` git trailer, I feel is useful for incremental adjustments while the others are helpful for when the issue is complete.

The next goal is to use the `linked_issues` to write a comment on those issues about the release that has been made (see #975). This would be broken with the use of `ref` especially if the issue was not auto-closed because the VCS does not interpret that, although you could argue that PSR currently supports a superset of closing git footers beyond what the team VCS's currently support individually. PSR's is also a bit more flexible for multiple issue specification that is not supported by most VCSs.

An alternative implementation?

Add ref and separate out the `linked_issues` from `closed_issues` on the commit object. Then, for #975, only write a release note on the `closed_issues` and use the Team VCS API to auto close the issue if it isn't already. The second part of this is to handle the superset problem when the Team VCS in use doesn't support that git trailer type but PSR does.

-----

Separately, you did not update the unit tests or documentation so don't mark off the PR completion checklist until you have.

JCHacking:
Hi
I understand your idea, to create two lists, one with linked issues and another with closed ones.

So, I understand that linked_issues should show all of them, and then closed_issues should only show those that would be closed.

codejedi365:
My response was more of a question of an alternative because your current solution would break the roadmap I had for the future. Are you trying to use linked_issues using the `refs` in your changelog?  Otherwise what is the intent for reading refs?

JCHacking:
Yes, I'm trying to read it because what I wanted was to put in each commit the task that had led to adding that change.

codejedi365:
Ok, so with that in mind likely `linked_issues` should be a superset of `closed_issues` that way for your changelog you are iterating through `linked_issues` for those references. But for the future auto closing of issues, then PSR would iterate through `closed_issues`

JCHacking:
Yeah, I think that would be the best way for customers to complete the changelog correctly and then be able to add that automatic closing functionality.

JCHacking:
I have uploaded the adapted tests and different types of refs that could be used.

Should I add closed_issues to this PR? Or should this go in another one?

codejedi365:
A separate PR likely makes sense for the closing issues part. Thank you.

JCHacking:
And so would the PR be valid, or would some change need to be made?

JCHacking:
Hi @codejedi365 
Do you need me to change anything else?

codejedi365:
Sorry I've been unavailable. I'll take a look this weekend.

JCHacking:
Hi @codejedi365 Is there a release date for this?

PR #1404, opened 2026-01-01
feat(changelog): Support remote git repositories as template sources

### LINKED ISSUE #1401: Support remote git repository as template source
### Description

Allow `template_dir` to reference a remote git repository instead of only local filesystem paths.

### Use cases

- Share changelog templates across multiple projects without copy-paste
- Enable an ecosystem of reusable changelog template repositories
- Simplify CI workflows by removing manual template download steps

### Possible implementation

Support a git URL syntax:

```toml
[tool.semantic_release.changelog]
template_dir = "git+https://github.com/user/changelog-templates.git@v1.0.0#templates"
```

Or explicit configuration:

```toml
[tool.semantic_release.changelog.template_source]
repo = "https://github.com/user/changelog-templates.git"
ref = "v1.0.0"  # tag, branch, or commit
path = "templates"  # subdirectory within repo
```

Using [fsspec](https://filesystem-spec.readthedocs.io/) with its git backend would abstract the storage layer, and support private repositories.

### Alternative solutions

- Download templates in CI before running semantic-release: Works but adds boilerplate to every project's workflow
- Git submodules: Adds unnecessary complexity for simple template sharing

bilelomrani1:
## Purpose

Closes #1401

Add support for remote repositories as template sources for changelog generation, allowing users to share templates across multiple projects without copy-paste or manual CI setup.

## Rationale

Currently, `template_dir` only accepts local filesystem paths, requiring users to either copy templates into each project, use git submodules or add manual download steps in CI workflows.

By leveraging [universal_pathlib](https://github.com/fsspec/universal_pathlib) which uses [fsspec](https://filesystem-spec.readthedocs.io/) under the hood, we can abstract the storage layer and support any fsspec-compatible backend with minimal code changes.

Security considerations:

- Chained protocols (e.g., `simplecache::file://`) are explicitly rejected to prevent bypass of path traversal checks
- Local path traversal validation is preserved for `file://`, `local://`, and bare paths
- Remote protocols skip local path validation since they cannot access the local filesystem

## How did you test?

1. Added tests covering:
   - `UPathLoader` template loading from memory filesystem
   - `StorageOptionsConfig` environment variable resolution
   - Chained protocol rejection
   - Valid `template_dir` acceptance for various protocols
   - Path traversal checks for local vs remote protocols
   - `recursive_render` with `UPath` sources

2. Validated end-to-end with a real GitHub-hosted template repository:

   ```toml
   [tool.semantic_release.changelog]
   template_dir = "github://bilelomrani1:psr-conventional-changelog-template@main/src/psr_templates/templates"
   ```

   Successfully generated changelog using remote templates.

## How to Verify

1. Install dependencies: `pip install -e .`
2. Create a test repository with a `pyproject.toml`:

   ```toml
   [tool.semantic_release.changelog]
   template_dir = "github://owner:repo@ref/path/to/templates"
   ```

4. Run `semantic-release changelog`
5. Verify templates are fetched from the remote repository

---

## PR Completion Checklist

- [x] Reviewed & followed the [Contributor Guidelines](https://python-semantic-release.readthedocs.io/en/stable/contributing/contributing_guide.html)

- [x] Commits follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard

- [ ] Appropriate End-to-End tests added/updated

- [x] Appropriate Documentation added/updated and syntax validated for sphinx build

### DIFF
--- a/src/semantic_release/cli/config.py
+++ b/src/semantic_release/cli/config.py
@@ -29,6 +29,7 @@ from pydantic import (
     model_validator,
 )
 from typing_extensions import Annotated, Self
+from upath import UPath
 from urllib3.util.url import parse_url
 
 import semantic_release.hvcs as hvcs
@@ -149,6 +150,39 @@ class DefaultChangelogTemplatesConfig(BaseModel):
         return self
 
 
+class StorageOptionsConfig(BaseModel):
+    """
+    fsspec storage options with environment variable support.
+
+    This configuration allows passing authentication and other options to fsspec
+    for remote template directories. Values can be either strings or EnvConfigVar
+    objects to resolve from environment variables.
+
+    Example configuration:
+        [tool.semantic_release.changelog.storage_options]
+        key = { env = "AWS_ACCESS_KEY_ID" }
+        secret = { env = "AWS_SECRET_ACCESS_KEY" }
+    """
+
+    model_config = {"extra": "allow"}
+
+    @model_validator(mode="before")
+    @classmethod
+    def resolve_env_vars(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        resolved: Dict[str, Any] = {}
+        for key, value in values.items():
+            if isinstance(value, dict) and "env" in value:
+                # This is an EnvConfigVar-like dict
+                env_var = EnvConfigVar(**value)
+                resolved_value = env_var.getvalue()
+                # Only include if we got a value
+                if resolved_value is not None:
+                    resolved[key] = resolved_value
+            elif value is not None:
+                resolved[key] = value
+        return resolved
+
+
 class ChangelogConfig(BaseModel):
     # TODO: BREAKING CHANGE v11, move to DefaultChangelogTemplatesConfig
     changelog_file: str = ""
@@ -162,6 +196,20 @@ class ChangelogConfig(BaseModel):
     mode: ChangelogMode = ChangelogMode.UPDATE
     insertion_flag: str = ""
     template_dir: str = "templates"
+    storage_options: StorageOptionsConfig = Field(default_factory=StorageOptionsConfig)
+
+    @field_validator("template_dir", mode="after")
+    @classmethod
+    def validate_template_dir(cls, val: str) -> str:
+        # Chained protocols (e.g., simplecache::s3://) are not supported.
+        # Attackers could use arbitrarily complex nested protocols to bypass
+        # local path security heuristics. For example, simplecache::file:///etc/passwd
+        # would have protocol "simplecache", not "file", bypassing the local path
+        # traversal check. Rather than trying to parse and validate nested chains,
+        # we disallow them entirely.
+        if "::" in val:
+            raise ValueError("Chained protocols are not supported for template_dir.")
+        return val
 
     @field_validator("exclude_commit_patterns", mode="after")
     @classmethod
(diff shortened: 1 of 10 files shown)


bilelomrani1:
While working on this feature, I noticed a potential limitation worth discussing for a future enhancement.

When using user templates, `recursive_render()` mirrors the template directory structure directly to the project root. For 
example: `template_dir/docs/CHANGELOG.md.j2`  →  `./docs/CHANGELOG.md`

With remote template sources, users may want to use a template from an external repository or storage that has its own 
directory structure. Currently, they cannot control where the rendered output lands in their project, it's dictated by the 
template's internal layout.

Users can already control which part of a remote template to use by pointing `template_dir` to a specific subdirectory:
```toml
template_dir = "git+https://github.com/org/templates@main/changelogs"
```

But, there's no way to render templates to a location other than the project root. This matters when users want their changelog in a subdirectory like `docs/` or `build/`, or in a more complex location, for instance in a monorepo.

I'm proposing to add a `changelog.output_dir` config option:
```toml
[tool.semantic_release.changelog]
template_dir = "git+https://github.com/org/templates@main"
output_dir = "docs/"
```
This would render the template rooted at `templates` on the remote, inside the `docs/` directory locally. Default would be `.` for backwards compatibility.

I wanted to flag it as a potential follow-up. Happy to hear thoughts on whether this is a use case worth supporting and which approach would fit best with the existing design.

bilelomrani1:
Just submitted a proposal that would address the previous limitation here: #1406

bilelomrani1:
Just a note: this is one of the smaller, more self-contained features from my recent batch of submissions. It adds support for loading templates from remote git repositories, which is useful for sharing templates across multiple projects.

I don't think this conflicts with the v11 refactoring work, so it could potentially be reviewed independently whenever you have time. No rush see #1407 for the broader context on my use case and how this fits in.

codejedi365:
I agree this should not conflict. Thanks for putting the effort in here as it would be awhile before I would have gotten to this when it was asked for before.

src/semantic_release/cli/config.py
```diff
@@ -162,6 +196,20 @@ class ChangelogConfig(BaseModel):
...
+    def validate_template_dir(cls, val: str) -> str:
+        # Chained protocols (e.g., simplecache::s3://) are not supported.
+        # Attackers could use arbitrarily complex nested protocols to bypass
+        # local path security heuristics. For example, simplecache::file:///etc/passwd
+        # would have protocol "simplecache", not "file", bypassing the local path
+        # traversal check. Rather than trying to parse and validate nested chains,
+        # we disallow them entirely.
+        if "::" in val:
+            raise ValueError("Chained protocols are not supported for template_dir.")
+        return val
```
bilelomrani1: We can instead check for explicit `"local://"`, `"file://"`, and bare paths to prevent accidental local filesystem access outside the project's repo. This catches common mistakes. This would allow chained protocols instead of disallowing them, which avoids confusing users with a seemingly arbitrary restriction.

I also just realized one workaround around this protection: `fsspec` allows users to register [custom protocols](https://filesystem-spec.readthedocs.io/en/latest/developer.html#implementing-a-backend) that could wrap the local filesystem. We cannot detect these, but my assumption is that this is not a real attack vector: it requires the user to deliberately install and configure such a protocol on their system. Users who do this are explicitly opting into local filesystem access, which is outside our threat model.

My assumption with this path traversal check is: our goal is to prevent accidental misconfiguration and obvious attacks, not to sandbox users from their own deliberate actions. 

cc @codejedi365, can you confirm how reasonable this assumption sounds to you?

PR #1444, opened 2026-05-09
fix: use resolve() instead of absolute() in verify_git_repo_dir

### LINKED ISSUE #1418: False "Found .git/ in higher parent directory" warning in monorepo setup
### Description

When following the [monorepo documentation](https://python-semantic-release.readthedocs.io/en/latest/configuration/configuration-guides/monorepos.html), I configure `repo_dir = ".."` to point from my package directory to the repository root. PSR then emits this warning:

```
WARNING:root:Found .git/ in higher parent directory rather than provided in configuration.
```

The warning is misleading because the configuration is correct—`..` does resolve to the repository root.

### Cause

In `semantic_release/cli/config.py`, the `verify_git_repo_dir` validator compares paths without resolving them:

```python
if dir_path.absolute() != found_path:
    logging.warning(...)
```

`Path("..").absolute()` returns `/path/to/package/..` rather than `/path/to/repo`, so the comparison fails even when both paths point to the same directory.

### Suggested Fix

```python
if dir_path.resolve() != found_path.resolve():
```

### Reproduction

Directory structure:
```
my-monorepo/           # .git/ here
├── package-a/
│   └── pyproject.toml  # repo_dir = ".."
```

Commands:
```bash
cd package-a
semantic-release version --print
# WARNING:root:Found .git/ in higher parent directory...
```

### Documented Workflow

The [monorepo guide](https://python-semantic-release.readthedocs.io/en/latest/configuration/configuration-guides/monorepos.html) instructs users to run PSR from the package subdirectory:

> "To release a package, change your current working directory to the package directory and execute PSR's `semantic-release version`."

In this workflow, the `.git/` directory will always be in a parent directory. The warning contradicts the documented approach.

## Environment

- python-semantic-release: 10.5.3
- Python: 3.12
- OS: Linux

aconal-com:
In monorepo setups where `repo_dir='..'` is configured for a subpackage, `Path('..').absolute()` returns `/path/to/package/..` rather than the resolved `/path/to/repo`. Using `resolve()` properly handles symlinks and normalizes paths, preventing false 'Found .git/ in higher parent directory' warnings.

Fixes python-semantic-release/python-semantic-release#1418

### DIFF
--- a/docs/configuration/configuration.rst
+++ b/docs/configuration/configuration.rst
@@ -1208,6 +1208,25 @@ from the :ref:`remote.name <config-remote-name>` location of your git repository
 
 ----
 
+.. _config-repo_dir:
+
+``repo_dir``
+""""""""""""
+
+**Type:** ``str``
+
+Specify the directory of the Git repository. This is used to determine the location of
+the repository for various operations, such as creating tags and commits. PSR will attempt
+to automatically determine the location of the repository in parent directories of the
+current working directory but will emit a warning if it is higher than this defined value.
+
+If a git repository is not found from this location or any parent directories, PSR will
+raise an error and exit.
+
+**Default:** ``"."``
+
+----
+
 .. _config-tag_format:
 
 ``tag_format``
--- a/docs/configuration/configuration.rst
+++ b/docs/configuration/configuration.rst
@@ -120,6 +120,52 @@ sure to use the correct root key depending on the configuration format you are u
 
 ----
 
+.. _config-add_partial_tags:
+
+``add_partial_tags``
+""""""""""""""""""""
+
+**Type:** ``bool``
+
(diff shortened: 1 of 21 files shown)


codejedi365:
@aconal-com, Thank you for the contribution. A few tips for contributing:

- make sure that your suggested test case not only succeeds after your fix but it should also fail without your fix. In this case it didn't fail which means the test case did not fully replicate your failure environment. Not to worry I figured it out but it will take longer to get your PRs merged if this happens.

- Make sure to use the PR template that is provided when submitting PRs.

codejedi365:
### :tada: This PR has been published as part of v10.6.1 :tada:

You can find more information about this release on the [GitHub Releases](https://github.com/python-semantic-release/python-semantic-release/releases/tag/v10.6.1) page.

